PR #1, opened 2020-06-23
Initial upload and migration from interns' old repo

hollyyuqizheng:
This PR contains code that the interns were working on for the past day in our old repo. This PR represents the master branch in our old repo, changes in which all three of us have looked at and approved. 

We have the initial design for the prototype's UI:

Screenshot 2020-06-23 at 4 30 19 PM

- The user can add new calendar events. When they hit the submit button, the new event will be displayed as a list on the UI.
- These events are also sent to the Calendar servlet. Right now, the new events are simply added to a List of calendar events.
- On the UI, if the user inputs an end time that is earlier than the start time, a warning message shows on the UI, and the event cannot be added.

Screenshot 2020-06-23 at 2 33 41 PM


jamesnam-google (changes requested):
For the capstone project, pod-mates should review each other's code first. Once they've approved, you can send the pull request to the hosts for final review.

raulcruise (approved):

talvarez0507 (approved):

portfolio/src/main/java/com/google/sps/data/CalendarEvent.java
```diff
@@ -0,0 +1,34 @@
+package com.google.sps.data;
+
+/** Models a calendar event that the user creates. */
```
jamesnam-google: Do you plan to use this data model for imported calendar events? In that case, this comment may be too restrictive. Consider updating the comment to,

Data model for a calendar event. It can be used for both creation and import flows.

portfolio/src/main/java/com/google/sps/data/CalendarEvent.java
```diff
@@ -0,0 +1,34 @@
...
+public class CalendarEvent {
+  private final String name;
+  private final String startTime;
```
jamesnam-google: Let's indicate the time unit. Ex: startTimeMillis. Also, the type is string. I don't think the type should be String. At least it should be long to store the timestamp millis.

Also, consider using the Instant Java class:
https://docs.oracle.com/javase/8/docs/api/java/time/Instant.html

It might be a good time to use the correct data types, so you don't have to refactor later.

portfolio/src/main/java/com/google/sps/data/CalendarEvent.java
```diff
@@ -0,0 +1,34 @@
...
+  private final String endTime;
+
+  // TODO: does an event need an ID?
```
jamesnam-google: Let's have an assignee here.

ex:
// TODO(holly...): Consider adding event ID if necessary.

portfolio/src/main/java/com/google/sps/data/Task.java
```diff
@@ -0,0 +1,39 @@
+package com.google.sps.data;
+
+/** Models a task that the user creates. */
```
jamesnam-google: Similarly, consider updating this comment so you can use it for both creation and import flows.

portfolio/src/main/java/com/google/sps/data/Task.java
```diff
@@ -0,0 +1,39 @@
...
+public final class Task {
+
+  private final String name;
+  private final String description;
+  private final Integer durationMinute;
```
jamesnam-google: Consider using the Duration class.
https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html

Again, it might be a good idea to start with the right data types, so you don't have to refactor later.

jamesnam-google: Will all tasks have the associated duration? If not, consider making this field optional.

portfolio/src/main/java/com/google/sps/data/Task.java
```diff
@@ -0,0 +1,39 @@
...
+  private final Integer priority;
+
+  // TODO: does a task need an ID?
```
jamesnam-google: Similarly, let's have an assignee.

portfolio/src/main/java/com/google/sps/data/Task.java
```diff
@@ -0,0 +1,39 @@
+package com.google.sps.data;
+
+/** Models a task that the user creates. */
+public final class Task {
+
+  private final String name;
+  private final String description;
+  private final Integer durationMinute;
+  private final Integer priority;
```
jamesnam-google: I would suggest that you introduce a TaskPriority class. That way, for example, you can enforce range.

portfolio/src/main/java/com/google/sps/servlets/CalendarServlet.java
```diff
@@ -0,0 +1,33 @@
...
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet that handles calendar event creation. */
+@WebServlet("/calendarServlet")
+public class CalendarServlet extends HttpServlet {
+
+  public List<CalendarEvent> eventList;
```
jamesnam-google: Consider simplifying the variable name to "events"

portfolio/src/main/webapp/calendar.js
```diff
@@ -0,0 +1,50 @@
+class CalendarEvent{
+  constructor(name, startTime, endTime) {
```
jamesnam-google, line 17: Let's indicate the time unit in the variable names.

ex: startTimeMillis, endTimeMillis

Also, you may want to start thinking about time zone, so that you don't have to do a huge refactor later. 

One approach you can consider is using the Date Javascript objects, as opposed to these timestamps. If we pass around the timestamps, then it can be difficult to reason about time zone, time unit, etc.

portfolio/src/main/webapp/calendar.js
```diff
@@ -0,0 +1,50 @@
+class CalendarEvent{
```
jamesnam-google: 1. I think there should be a space before {. Can we use some tool that can automatically detect these style issues?
2. Can we have some comment about this class?
https://google.github.io/styleguide/jsguide.html#formatting-comments

portfolio/src/main/webapp/calendar.js
```diff
@@ -0,0 +1,50 @@
...
+}
+
+/** 
+ * Onclick function for "Add this event" button.
+ * Retrieves the new event's name, start and end times, and creates a new 
+ * CalendarEvent from this information. Adds the new event to the 
+ * event list. 
+ */
+function createNewCalendarEvent() {
+  const startTime = document.getElementById('new-event-start-time').value;
```
jamesnam-google: 1. Let's try to indicate time unit in the variable names.
2. Let's try to use Date objects wherever possible.

portfolio/src/main/java/com/google/sps/servlets/CalendarServlet.java
```diff
@@ -0,0 +1,33 @@
...
+import com.google.sps.data.CalendarEvent;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet that handles calendar event creation. */
```
jamesnam-google: From what I can tell, currently the client can create calendar events on the fly through the server. I think it can later cause an issue where it's difficult to reconcile the client state vs server state.

I would recommend that you maintain the states on the client, until the schedule computation happens. This is also worthwhile to cover in the design doc. What kind of APIs will the client invoke to the server?

Consider the following approach, for example.
1. Suppose the user created a calendar event. At this point, client can just hold onto the event, and not tell the sever yet.
2. Similarly, client can create tasks that are kept track on the client side.
3. Now when the user tries to compute the schedule, client can call some API that takes the tasks and events, and returns the computed schedule.

In other words, while keeping track of these intermediate states on the server can be a good idea, it brings in complexity. And at this stage you may want to avoid keeping track of states on both the client and the server. Consider making the server operations state-less at this time.

portfolio/src/main/webapp/task.js
```diff
@@ -0,0 +1,8 @@
+class Task{
```
jamesnam-google: Let's have a space before {. And let's add some comment on this class.

portfolio/src/main/webapp/task.js
```diff
@@ -0,0 +1,8 @@
...
+  constructor(name, description, durationMinute, priority) {
```
jamesnam-google: durationMinute => durationMinutes. Also, the comment should indicate which fields are optional. And it should indicate how the optional status is conveyed (null? undefined?)

portfolio/target/portfolio-1/calendar.js
```diff
@@ -0,0 +1,51 @@
+class CalendarEvent{
```
jamesnam-google: Similar comment on comment, space before {

portfolio/target/portfolio-1/calendar.js
```diff
@@ -0,0 +1,51 @@
...
+  constructor(name, startTime, endTime) {
```
jamesnam-google, line 13: Similar comment about using Date instead of timestamp variables.

portfolio/target/portfolio-1/script.js
```diff
@@ -0,0 +1,28 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * Adds a random greeting to the page.
+ */
+function addRandomGreeting() {
```
jamesnam-google: Is this random greeting necessary? Maybe you want to add some fun features, but consider avoiding adding these extraneous functionalities at this stage.

portfolio/target/portfolio-1/style.css
```diff
@@ -0,0 +1,4 @@
+#event-end-time-warning {
+  /* default visibility is hidden */
```
jamesnam-google: I don't think this comment is necessary.

portfolio/target/portfolio-1/task.js
```diff
@@ -0,0 +1,8 @@
+class Task{
```
jamesnam-google: Similar comment about space before {, comment

jamesnam-google (changes requested):

jamesnam-google:
Please see my comments. Also, consider adding some basic unit test coverage for the Java code change. You'll want to have unit test in each pull request that involves logic change.

If you have extra time, you can explore how you might be able to add unit test for the Javascript change as well. But for now I'm okay with first unit testing in Java.

portfolio/src/main/java/com/google/sps/data/CalendarEvent.java
hollyyuqizheng: Updated. The class should be used for both created and imported events.

portfolio/src/main/java/com/google/sps/data/CalendarEvent.java
hollyyuqizheng: Updated.

portfolio/src/main/java/com/google/sps/data/Task.java
hollyyuqizheng: Updated.

portfolio/src/main/java/com/google/sps/data/Task.java
hollyyuqizheng: Updated.

portfolio/src/main/java/com/google/sps/servlets/CalendarServlet.java
hollyyuqizheng: Updated.

portfolio/src/main/webapp/calendar.js
hollyyuqizheng: Updated. Now I am using the Date object, to build startTime and endTime.

portfolio/src/main/webapp/calendar.js
hollyyuqizheng: Updated.

portfolio/src/main/webapp/calendar.js
hollyyuqizheng: Updated to date object.

portfolio/target/portfolio-1/script.js
hollyyuqizheng: I think the original script.js was an example file from the intern team's clean repo, and this file was generated during maven testing. I forgot to delete the script.js file here under target folder, so yes this file is unnecessary.

portfolio/target/portfolio-1/style.css
hollyyuqizheng: Updated.

portfolio/src/main/java/com/google/sps/servlets/CalendarServlet.java
hollyyuqizheng: Updated. I created an array on the client side that stores each new calendar event as it is created by the user, or imported when I later implement that feature. When the user clicks on the "Start scheduling" button, this array of calendar events is sent to the calendar servlet all at once.

portfolio/src/main/java/com/google/sps/data/CalendarEvent.java
hollyyuqizheng: In the calendar servlet, I was hoping to use the fromJson method from the Gson library to unpack a JSON String directly into an instance of the CalendarEvent class. It seems easier for this method to work if the time parameters are of type String. The workaround I have right now is, the start and end times parameters are of type String, but the getter methods will use Instant.parse and return an instance of Instant class.

portfolio/src/main/java/com/google/sps/data/Task.java
raulcruise: Done.

raulcruise: All tasks will have a duration so we have a time to fit within the calendar.

portfolio/src/main/java/com/google/sps/data/Task.java
raulcruise: Done.

portfolio/src/main/webapp/task.js
raulcruise: Done.

portfolio/pom.xml
```diff
@@ -0,0 +1,76 @@
...
+          <execution>
+            <id>install-formatter-hook</id>
+            <goals>
+              <goal>install-hooks</goal>
+            </goals>
+          </execution>
+          <execution>
+            <id>validate-code-format</id>
+            <goals>
+              <goal>validate-code-format</goal>
```
jamesnam-google, line 69: Did you have a chance to explore using eslint to format frontend changes as well?

portfolio/src/main/java/com/google/sps/data/CalendarEvent.java
```diff
@@ -0,0 +1,39 @@
+package com.google.sps.data;
+
+import java.time.Instant;
+
+/** Models a calendar event. It can be used for both creation and import flow. */
+public class CalendarEvent {
+  private String name;
+  private final String startTimeString;
```
jamesnam-google: I think these variables should be Instant. You can perform the conversion in the constructor (that takes in String's).

portfolio/src/main/java/com/google/sps/data/CalendarEvent.java
```diff
@@ -0,0 +1,39 @@
...
+  private final String endTimeString;
+
+  // TODO(hollyyuqizheng): Add an ID field if necessary.
+
+  public CalendarEvent(String name, String startTimeString, String endTimeString) {
```
jamesnam-google, line 22: Let's add some comment. And it should describe what startTimeString and endTimeString look like.

portfolio/src/main/java/com/google/sps/data/CalendarEvent.java
```diff
@@ -0,0 +1,39 @@
...
+    }
+    if (endTimeString == null) {
+      throw new IllegalArgumentException("Event needs an end time");
+    }
+    this.name = name;
+    this.startTimeString = startTimeString;
+    this.endTimeString = endTimeString;
+  }
+
+  public String getName() {
```
jamesnam-google, line 39: Is this valid? I didn't see a comment above that indicates which fields are optional.

portfolio/src/main/java/com/google/sps/data/CalendarEvent.java
```diff
@@ -0,0 +1,39 @@
...
+    if (name == null) {
+      name = "New Event";
```
jamesnam-google: The getter should not have a side effect. I think this getter should just return name. And the method should explain whether it can be null or not.

In my opinion, name, start time, end time seem like necessary fields.

portfolio/src/main/java/com/google/sps/data/CalendarEvent.java
```diff
@@ -0,0 +1,39 @@
+package com.google.sps.data;
+
+import java.time.Instant;
+
+/** Models a calendar event. It can be used for both creation and import flow. */
+public class CalendarEvent {
+  private String name;
```
jamesnam-google, line 7: Please indicate which fields are optional and which are required. And indicate which fields can contain null.

portfolio/src/main/java/com/google/sps/data/Task.java
```diff
@@ -0,0 +1,51 @@
+package com.google.sps.data;
+
+import java.time.Duration;
+
+/** Models a task. It can be used for both creation and import flow. */
+public final class Task {
+
+  private final String name;
```
jamesnam-google, line 12: Similarly, please indicate which fields are optional and which fields can store null.

Consider using java.util.Optional for fields that can be optional.

portfolio/src/main/java/com/google/sps/data/Task.java
```diff
@@ -0,0 +1,51 @@
...
+  private final String description;
+  private final Duration durationMinute;
```
jamesnam-google: Now that we're using the Duration class, we can omit the unit.

durationMinute => duration

portfolio/src/main/java/com/google/sps/data/Task.java
```diff
@@ -0,0 +1,51 @@
...
+  private final TaskPriority priority;
+
+  // TODO(raulcruise): Add an ID field if necessary.
+
+  // The constructor will make sure that all necessary parameters are passed in,
+  // and populate each class variable appropriately.
+  public Task(String name, String description, Duration durationMinute, Integer priority) {
```
jamesnam-google: Let's use javadoc style.

/**
 * ...
 * ...
 */

Also, please indicate in the constructor which fields can be optional or null. If not, clarify that none of the fields should be null.

jamesnam-google: I would take TaskPriority instead of Integer. Otherwise, caller can pass a value like -100, which I assume is invalid.

And as I mentioned above, we can rename durationMinute to duration.

portfolio/src/main/java/com/google/sps/data/TaskPriority.java
```diff
@@ -0,0 +1,17 @@
+package com.google.sps.data;
+
+/** The TraskPriority class handles range checking, and allows retrieval of the priority value. */
+public final class TaskPriority {
+  private final int priority;
+
+  TaskPriority(int priority) {
+    if (priority >= 1 || priority <= 5) {
```
jamesnam-google: This case is valid. I think you need to negate. And it'd be nice to add some unit test on this class.

portfolio/src/main/java/com/google/sps/data/Task.java
```diff
@@ -0,0 +1,51 @@
...
+
+  public String getName() {
+    return name;
+  }
+
+  public String getDescription() {
+    return description;
+  }
+
+  public Duration getDurationMinute() {
```
jamesnam-google: getDuration()

portfolio/src/main/java/com/google/sps/servlets/CalendarServlet.java
```diff
@@ -0,0 +1,37 @@
...
+
+  public List<CalendarEvent> events;
+
+  public CalendarServlet() {
+    events = new ArrayList<CalendarEvent>();
+  }
+
+  /**
+   * Retrieves the new calendar event's information from the POST request. Creates a new instance of
+   * the CalendarEvent class and adds it to a list of calendar events.
```
jamesnam-google: I still see that this servlet is stateful. For example, consider the scenario below.
1. User A visits the web app, and invokes this servlet method with their (private) calendar events.
2. User B does the same with their own private calendar events.

Then this servlet will end up storing both users' calendar events, right? So I think it'd be simpler/safer to make the servlet stateless.

For example, consider introducing ScheduleTasksServlet, which takes in the user's tasks (to complete) and existing events, as well as preferences etc, returns the scheduling output. It does not have to update some local variable.

portfolio/src/main/webapp/calendar.js
```diff
@@ -0,0 +1,82 @@
+/** 
+ * Models a calendar event that is displayed on the UI.
+ * This class is useful when the event information needs to be sent 
+ * to the calendar servlet through a POST request. 
+ * This class contains all the event information that can be easily
+ * converted into a JSON string. 
+ * All fields are required. 
+ */
+class CalendarEvent {
```
jamesnam-google, line 8: Indicate in the class method the types of the inputs.

jamesnam-google (changes requested):
Please see my comments, especially around ensuring that server doesn't keep track of states from different requests. For now, I strongly recommend that servlets have no side effects.

portfolio/pom.xml
talvarez0507: I'm looking into it currently. I made a new branch called "lint-checks" which has code based on the links you provided. We ran into some errors when testing locally so I have not made a pull request yet, but once we know it definitely works for everyone I will make the PR.

portfolio/src/main/java/com/google/sps/data/Task.java
raulcruise: Done.

portfolio/src/main/java/com/google/sps/data/Task.java
raulcruise: Done.

portfolio/src/main/java/com/google/sps/data/Task.java
raulcruise: Done.

raulcruise: Done.

portfolio/src/main/java/com/google/sps/data/CalendarEvent.java
hollyyuqizheng: I see. Updated accordingly.

portfolio/src/main/java/com/google/sps/data/CalendarEvent.java
hollyyuqizheng: Updated.

portfolio/src/main/webapp/calendar.js
hollyyuqizheng: Updated.

portfolio/src/main/java/com/google/sps/data/TaskPriority.java
raulcruise: Done.

portfolio/src/main/java/com/google/sps/data/Task.java
raulcruise: Done.

portfolio/src/main/java/com/google/sps/data/CalendarEvent.java
hollyyuqizheng: Updated in constructor comment.

portfolio/src/main/java/com/google/sps/data/CalendarEvent.java
hollyyuqizheng: Updated. I am making all three fields mandatory now, and I added the error checking for null name in the constructor.

portfolio/src/main/java/com/google/sps/servlets/CalendarServlet.java
hollyyuqizheng: I see. I originally built the Calendar servlet thinking that there will be some backend functionalities that only apply to the calendar part, but based on the goal of making the servlet stateless, I don't think it's necessary to have a separate servlet for the calendar anymore. I think the servlet that handles taking in the tasks and events and outputting the schedule is something Tomas is building already, so in this PR, I will delete the unnecessary CalendarServlet.java file for now. If it turns out that Tomas had other plans of incorporating the scheduling algorithm with a servlet, I will revisit this with him. 

As of right now, the calendar's client-side only needs to send a list of calendar events to the ScheduleTaskServlet, so having an additional CalendarServlet is not necessary.

portfolio/target/portfolio-1/calendar.js
hollyyuqizheng: Updated, although in future PR these files under target/ will be ignored.

portfolio/pom.xml
hollyyuqizheng: I just configured eslint and ran files through it.

portfolio/src/main/java/com/google/sps/data/CalendarEvent.java
```diff
@@ -0,0 +1,49 @@
...
+  private final Instant startTimeString;
```
jamesnam-google: Now that we're using Instant, this variable can be startTime

portfolio/src/main/java/com/google/sps/data/CalendarEvent.java
```diff
@@ -0,0 +1,49 @@
...
+  private final Instant endTimeString;
```
jamesnam-google: Same here. endTimeString => endTime

portfolio/src/main/java/com/google/sps/data/CalendarEvent.java
```diff
@@ -0,0 +1,49 @@
...
+
+  // TODO(hollyyuqizheng): Add an ID field if necessary.
+
+  /**
+   * Constructs a calendar event.
+   *
+   * @param name: Name for the event, can be null startTimeString: a string representation of the
```
jamesnam-google: Let's have one @param for each parameter

jamesnam-google: The comment here says name can be null, but the logic below checks for non-null.

portfolio/src/main/java/com/google/sps/data/TaskPriority.java
```diff
@@ -0,0 +1,18 @@
+package com.google.sps.data;
+
+/** Handles range checking, and allows retrieval of the priority value. */
+public final class TaskPriority {
```
jamesnam-google, line 4: Consider adding some static variables.

public static int MIN_PRIORITY = 1;
public static int MAX_PRIORITY = 5;

Then you can use them in the constructor.

portfolio/src/main/webapp/calendar.js
```diff
@@ -0,0 +1,79 @@
+/**
+ * Models a calendar event that is displayed on the UI.
+ * This class is useful when the event information needs to be sent
+ * to the calendar servlet through a POST request.
+ * This class contains all the event information that can be easily
+ * converted into a JSON string.
+ * All fields are required.
+ * @param name: of type String, name of event
```
jamesnam-google: The @param comment can be a method comment on the constructor, rather than a class level comment.

Example:
https://google.github.io/styleguide/jsguide.html#jsdoc-tags

jamesnam-google: And let's have one @param for each parameter.

portfolio/pom.xml
jamesnam-google: Thank you!

portfolio/src/main/webapp/task.js
```diff
@@ -0,0 +1,17 @@
+/**
+ * Models a task that is displayed on the UI.
+ * This class is useful when the task information needs to be sent
+ * to the task servlet through a POST request.
+ * This class contains all the task information that can be easily
+ * converted into a JSON string.
+ * All fields are required except description, which will be null
```
jamesnam-google: This can be a method comment on the constructor.

Example:
https://google.github.io/styleguide/jsguide.html#jsdoc-tags

portfolio/src/main/webapp/task.js
```diff
@@ -0,0 +1,17 @@
...
+ * when not set.
+ */
+class Task {
+  constructor(name, description, durationMinutes, priority) {
```
jamesnam-google: In the method comment for the constructor let's indicate the type of the parameters.

portfolio/src/test/java/com/google/sps/CalendarEventTest.java
```diff
@@ -0,0 +1,29 @@
...
+
+@RunWith(JUnit4.class)
+public final class CalendarEventTest {
+
+  /**
+   * Testing calendar event initialization and conversion between String and Instant for the start
+   * and end times.
+   */
+  @Test
+  public void TestTimeStringConversion() {
```
jamesnam-google: Test method usually begins with lower-case.

ex:
public void testTimeStringConversion()

portfolio/src/test/java/com/google/sps/TaskPriorityTest.java
```diff
@@ -0,0 +1,70 @@
...
+import com.google.sps.data.TaskPriority;
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public final class TaskPriorityTest {
+  private static final int MAX_ACCEPTABLE_VALUE = 5;
+  private static final int MIN_ACCEPTABLE_VALUE = 1;
```
jamesnam-google: As I mentioned earlier, these can be public constants in TaskPriority.

jamesnam-google (approved):
Approving the changes with comments.

rgcase (approved):

portfolio/src/main/java/com/google/sps/data/TaskPriority.java
hollyyuqizheng: Updated.

portfolio/src/main/webapp/task.js
raulcruise: Done.

portfolio/src/test/java/com/google/sps/TaskPriorityTest.java
raulcruise: Done.

portfolio/src/main/java/com/google/sps/data/TaskPriority.java
raulcruise: Done.

PR #2, opened 2020-06-25
Request data handling

talvarez0507:
This PR implements the basic classes that will deal with the data for scheduling requests. These classes just hold events, tasks, start of day, end of day in one place.


portfolio/src/main/java/com/google/sps/data/ScheduleRequest.java
```diff
@@ -0,0 +1,69 @@
...
+      throw new IllegalArgumentException("Request needs an end time");
+    }
+    this.events = events;
+    this.tasks = tasks;
+    // Converts time from string representation into an instance of Instant class.
+    this.startTime = Instant.parse(startTimeString);
+    this.endTime = Instant.parse(endTimeString);
+  }
+
+  /* Because all three fields are required, the following getters won't return null. */
```
hollyyuqizheng: nit: three => four fields

portfolio/src/main/webapp/request.js
```diff
@@ -0,0 +1,16 @@
+/**
+ * Models the requests sent to the algorithm for scheduling. This class is
+ * useful when the request information needs to be sent to the scheduling
+ * servlet through a POST request. This class contains all the request
+ * information that can be easily converted into a JSON string.
+ */
+class ScheduleRequest {
+  constructor(events, tasks, startTime, endTime) {
```
hollyyuqizheng: Would be nice to include comments on each parameter's type in the constructor.

portfolio/src/main/java/com/google/sps/data/ScheduleRequest.java
```diff
@@ -0,0 +1,69 @@
+package com.google.sps.data;
+
+import java.time.Instant;
+import java.util.Collection;
+/**
```
hollyyuqizheng: You might need to import CalendarEvent and Task?

portfolio/src/main/java/com/google/sps/data/ScheduleRequest.java
```diff
@@ -0,0 +1,69 @@
...
+ * Models request that are sent to the algorithm. It can be used for both
+ * creation and import flow.
+ */
+public class ScheduleRequest {
+  private Collection<CalendarEvent> events;
+  private Collection<Task> tasks;
+  private final Instant startTime;
+  private final Instant endTime;
```
hollyyuqizheng: These times represent the start and end of working hours? If so, maybe a better name for the time variables could be "workHourStartTimeInstant"

hollyyuqizheng (changes requested):
Some small suggestions

portfolio/src/main/java/com/google/sps/data/ScheduleRequest.java
```diff
@@ -0,0 +1,72 @@
...
+  /**
+   * Constructs a request to be scheduled.
+   *
+   * @param events: events happening the day of the scheduling,
+   * @param tasks: the tasks a user would like to schedule on that day
+   * @param workHoursStartTimeString: a string representation of the work hours start
+   *     time. The string is in format Day Month Date Year HH:MM:SS GMT-Time-zone
+   * @param workHoursEndTimeString: a string representation of the work hours end time.
+   *     The string's format is the same as startTimeString's.
+   * All of these fields are required for a request (can be empty not null).
```
hollyyuqizheng: When will these fields be empty?

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
```diff
@@ -0,0 +1,116 @@
...
+   * working hours.
+   */
+  public static TimeRange[] emptyTimeRanges(
+      List<CalendarEvent> events, long startTimeMinutes, long endTimeMinutes) {
+    List<TimeRange> possibleTimes = new ArrayList<TimeRange>();
+    // This represents the earliest time that we can schedule a window for the
+    // meeting. As events are processed, this changes to their end times.
+    long earliestPossibleSoFar = startTimeMinutes;
+    for (CalendarEvent event : events) {
+      // Make sure that there is some time between the events and is it not
```
hollyyuqizheng: nit: "is it" => "it is"

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
```diff
@@ -0,0 +1,116 @@
...
+
+  private static final Comparator<Task> sortByDuration =
+      Comparator.comparing(Task::getDurationSeconds).thenComparing(Task::getName);
+
+  /**
+   * This method returns a Task Collection which the greedy algorithm
+   * determines should be scheduled for the person.
+   */
+  public Collection<Task> greedy(Collection<CalendarEvent> events, Collection<Task> tasks,
+      long startTimeMinutes, long endTimeMinutes) {
```
hollyyuqizheng: I would change the variable names to something like workHourStartTimeMinutes, just as you did in other classes, to keep things uniform.

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
```diff
@@ -0,0 +1,116 @@
...
+    TimeRange[] availableTimes = emptyTimeRanges(eventsList, startTimeMinutes, endTimeMinutes);
+
+    List<Task> scheduledTasks = new ArrayList<Task>();
+    // Index of TimeRange we are in
+    int i = 0;
+    // Index of Task we are trying to schedule
+    int j = 0;
+    // Integer indicating the time (in minutes) we are currently trying to
+    // schedule events in
+    long time = 0;
```
hollyyuqizheng: I would rename this to currentTime or something similar.

hollyyuqizheng (changes requested):
Some suggestions on style mostly.

portfolio/src/test/java/com/google/sps/FindScheduleTest.java
```diff
@@ -0,0 +1,136 @@
...
+            new Task("Task 3", "Third task", 45, LOW_PRIORITY, "2020-06-25T15:30:00Z"),
+            new Task("Task 6", "Sixth task", 45, LOW_PRIORITY, "2020-06-25T16:15:00Z"));
+
+    Assert.assertEquals(expected, actual);
+  }
+
+  // TODO(tomasalvarez): Test edge cases
+  // 1: Events out of the scheduling bounds
+  // 2: Scheduling time is more than 24 hours
+  // 3: ???
```
hollyyuqizheng: Maybe we could do something like only half of the tasks can be scheduled because there isn't enough free space for the day?

portfolio/src/test/java/com/google/sps/FindScheduleTest.java
```diff
@@ -0,0 +1,100 @@
...
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+/** */
```
raulcruise: Remove line or add a comment.

portfolio/src/main/java/com/google/sps/data/ScheduleRequest.java
talvarez0507: Done.

portfolio/src/main/webapp/request.js

portfolio/src/main/java/com/google/sps/data/ScheduleRequest.java
talvarez0507: Not necessary because of the package.

portfolio/src/main/java/com/google/sps/data/ScheduleRequest.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/data/ScheduleRequest.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
talvarez0507: Done.

portfolio/src/test/java/com/google/sps/FindScheduleTest.java
talvarez0507: I added a new TODO for this case, thank you for the suggestion.

portfolio/src/test/java/com/google/sps/FindScheduleTest.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/data/TimeRange.java
```diff
@@ -0,0 +1,79 @@
...
+  public long end() {
+    return start + duration;
+  }
+
+  @Override
+  public boolean equals(Object other) {
+    return other instanceof TimeRange && equals(this, (TimeRange) other);
+  }
+
+  @Override
```
hollyyuqizheng, line 55: I feel like we might end up not needing this hashCode function, so maybe it's okay to delete?

hollyyuqizheng (approved):

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
```diff
@@ -0,0 +1,116 @@
...
+        emptyTimeRanges(eventsList, workHoursStartTimeSeconds, workHoursEndTimeSeconds);
+
+    List<Task> scheduledTasks = new ArrayList<Task>();
+    // Index of TimeRange we are in
+    int i = 0;
+    // Index of Task we are trying to schedule
+    int j = 0;
+    // Integer indicating the start time (in minutes) we are currently trying to
+    // schedule events in
+    long scheduleTimeSeconds = 0;
```
raulcruise: The comment indicates that you're declaring a int in units "minutes" but you end up declaring a long int with the unit seconds.

portfolio/src/main/java/com/google/sps/data/ScheduleRequest.java
```diff
@@ -0,0 +1,75 @@
...
+  }
+  public Instant getWorkHoursStartTimeInstant() {
+    return workHoursStartTime;
+  }
+  public Instant getWorkHoursEndTimeInstant() {
+    return workHoursEndTime;
+  }
+
+  // These methods return the instant as a number of seconds from 00:00:00 UTC
+  // on 1 January 1970
```
raulcruise: Nit: period at end.

portfolio/src/main/webapp/request.js
```diff
@@ -0,0 +1,25 @@
+/**
+ * Models the requests sent to the algorithm for scheduling. This class is
+ * useful when the request information needs to be sent to the scheduling
+ * servlet through a POST request. This class contains all the request
+ * information that can be easily converted into a JSON string.
+ */
+class ScheduleRequest {
+    /**
```
raulcruise: Nit: Misaligned beginning to javadoc style format.

raulcruise (approved):

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/data/ScheduleRequest.java
talvarez0507: Done.

portfolio/src/main/webapp/request.js

portfolio/src/main/java/com/google/sps/data/CalendarEvent.java
```diff
@@ -47,4 +47,12 @@ public Instant getStartTimeInstant() {
   public Instant getEndTimeInstant() {
     return endTime;
   }
+
+  public long getStartTimeLong() {
```
rgcase: Here and below, I don't think the `Long` suffix is useful here. What would be more useful is if the suffix was the units instead, like `getStartTimeSeconds()`.

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
```diff
@@ -0,0 +1,116 @@
...
+
+public final class FindSchedule {
+  private static final Comparator<CalendarEvent> sortAscending =
+      Comparator.comparingLong(CalendarEvent::getStartTimeLong);
+
+  private static final Comparator<Task> sortByDuration =
+      Comparator.comparing(Task::getDurationSeconds).thenComparing(Task::getName);
+
+  /**
+   * This method returns a Task Collection which the greedy algorithm
```
rgcase: nit: The "This method..." is redundant, the reader knows that the comment refers to this method.

Also, it's not really obvious to me from the comment + method signature what this method is supposed to do. It looks like it's using a greedy algorithm to choose where each task should be scheduled based on existing calendar events (and working hours), but it just returns `Task`s which don't have a start and end time, so it doesn't seem to be actually choosing times for these. Does this just filter out ones that are impossible to schedule?

portfolio/src/main/java/com/google/sps/data/Task.java
```diff
@@ -13,32 +14,55 @@
...
   /**
    * The constructor will make sure that all necessary parameters are passed in, and populate each
    * class variable appropriately. None of the fields can be null with exception to the description
    * as it is optional. If the description is passed in as null then it will be stored as an empty
-   * Optional object.
+   * Optional object. Tasks by default do not have a scheduled time, they are only assigned this
+   * field as a result of getting scheduled by the scheduling algorithm.
    */
-  public Task(String name, String description, Duration duration, TaskPriority priority) {
+  public Task(String name, String description, long durationMinutes, TaskPriority priority) {
```
rgcase: I think we should try to stick with using `Duration` here. Why do we need to use `long` instead?

portfolio/src/main/java/com/google/sps/data/CalendarEvent.java
jamesnam-google: Consider clarifying the time unit in the method name.

ex: getStartTimeSecondsLong (or just getStartTimeSeconds)

portfolio/src/main/java/com/google/sps/data/CalendarEvent.java
```diff
@@ -47,4 +47,12 @@ public Instant getStartTimeInstant() {
...
+    return startTime.getEpochSecond();
+  }
+
+  public long getEndTimeLong() {
```
jamesnam-google: Same here

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
```diff
@@ -0,0 +1,116 @@
...
+
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.List;
+
+public final class FindSchedule {
```
jamesnam-google: Consider adding some class comment. Also, how do you think we can make this class extensible? Consider adding TODO comment if you think it can be made more extensible, as we add scheduling algorithms.

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
```diff
@@ -0,0 +1,116 @@
...
+  private static final Comparator<CalendarEvent> sortAscending =
```
jamesnam-google: I think the variable name should make it clear which field is used for sorting.

ex:
sortByEventStartTimeAscending

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
```diff
@@ -0,0 +1,116 @@
...
+      Comparator.comparingLong(CalendarEvent::getStartTimeLong);
+
+  private static final Comparator<Task> sortByDuration =
```
jamesnam-google: I think the variable name should make it clear that it uses task duration, then name.

ex:
sortByTaskDurationThenName

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
```diff
@@ -0,0 +1,116 @@
...
+      Comparator.comparing(Task::getDurationSeconds).thenComparing(Task::getName);
+
+  /**
+   * This method returns a Task Collection which the greedy algorithm
+   * determines should be scheduled for the person.
+   */
+  public Collection<Task> greedy(Collection<CalendarEvent> events, Collection<Task> tasks,
```
jamesnam-google: Let's try to have one param per line.

ex:
greedy(
    Collection..,
    Collection...,
    ....,
    ...);

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
```diff
@@ -0,0 +1,116 @@
...
+      long workHoursStartTimeSeconds, long workHoursEndTimeSeconds) {
```
jamesnam-google: I think we should try to use Instant as parameter wherever possible. Can this method take in Instant params?

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
jamesnam-google: Consider clarifying further in the method comment what you mean by greedy algorithm. For example does it try to fit in the task with the highest priority first?

portfolio/src/main/java/com/google/sps/data/Task.java
```diff
@@ -13,32 +14,55 @@
   private final Optional<String> description;
   private final Duration duration;
   private final TaskPriority priority;
+  private Instant scheduledTime;
```
rgcase: I don't really like how a `Task` will sometimes have a `scheduledTime` and sometimes not, and there's not really any way to know about it ahead of time. Consider instead creating a wrapper class like:
```
public class ScheduledTask {
  private final Task task;
  private final Instant startTime;

  ...
}
```

then you can statically know whether a `Task` has been scheduled or not.

portfolio/src/main/java/com/google/sps/data/Task.java
```diff
@@ -53,7 +77,33 @@ public Duration getDuration() {
...
+  // These methods provide a way to tell if two objects are both Task objects
+  // and have all the same fields
+
+  @Override
+  public boolean equals(Object other) {
+    return other instanceof Task && equals(this, (Task) other);
+  }
+
+  private static boolean equals(Task a, Task b) {
+    return a.name.equals(b.name) && a.description.equals(b.description)
```
rgcase: nit: Line break before each `&&`, don't mix lines with one clause and lines with multiple clauses.

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
```diff
@@ -0,0 +1,116 @@
...
+    return scheduledTasks;
+  }
+
+  /**
+   * This method returns an TimeRange array which represent the periods of
+   * time that are empty of events and lie completely inside the person's
+   * working hours.
+   */
+  public static TimeRange[] emptyTimeRanges(
+      List<CalendarEvent> events, long workHoursStartTimeSeconds, long workHoursEndTimeSeconds) {
```
jamesnam-google: Can we try to take in Instant objects rather than long?

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+    return scheduledTasks;
+  }
+
+  /**
+   * This method returns an TimeRange array which represent the periods of
+   * time that are empty of events and lie completely inside the person's
+   * working hours.
+   */
+  public static TimeRange[] emptyTimeRanges(
```
jamesnam-google: 1. Is there a reason why you return array instead of list? Lists are more flexible and easier to work with. Consider returning a list of TimeRange.
2. I think it'd be nice to clarify in the method name that it gets or computes the time ranges. So for example, computeEmptyTimeRanges or getEmptyTimeRanges.

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
```diff
@@ -0,0 +1,116 @@
...
+      List<CalendarEvent> events, long workHoursStartTimeSeconds, long workHoursEndTimeSeconds) {
+    List<TimeRange> possibleTimes = new ArrayList<TimeRange>();
+    // This represents the earliest time that we can schedule a window for the
+    // meeting. As events are processed, this changes to their end times.
+    long earliestPossibleSoFar = workHoursStartTimeSeconds;
```
jamesnam-google: The time unit is not clear in the variable name. And consider using Instant here as well.

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
```diff
@@ -0,0 +1,116 @@
...
+    for (CalendarEvent event : events) {
+      // Make sure that there is some time between the events and it is not
+      // later than the person's working hours ending time.
+      if (event.getStartTimeLong() - earliestPossibleSoFar > (long) 0
```
jamesnam-google: event.getStartTimeLong() - earliestPossibleSoFar > (long) 0

is equivalent to

event.getStartTimeLong() > earliestPossibleSoFar

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
```diff
@@ -0,0 +1,116 @@
...
+    while (i < availableTimes.length && j < tasksArray.length) {
+      TimeRange timeRange = availableTimes[i];
+      Task task = tasksArray[j];
+      // Either time is already past the start of the time range or we should
+      // update it (maybe this is our first iteration in the range).
+      scheduleTimeSeconds = Math.max(scheduleTimeSeconds, timeRange.start());
+      // The task can be scheduled in the current time range.
+      if (scheduleTimeSeconds + task.getDurationSeconds() <= timeRange.end()) {
+        Task scheduledTask;
+        if (task.getDescription().isPresent()) {
```
rgcase: This is kind of messy, but I think it'll go away if you start using a new `ScheduledTask` like I mentioned in a previous comment.

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
```diff
@@ -0,0 +1,116 @@
...
+      // later than the person's working hours ending time.
+      if (event.getStartTimeLong() - earliestPossibleSoFar > (long) 0
+          && event.getStartTimeLong() <= workHoursEndTimeSeconds) {
+        possibleTimes.add(TimeRange.fromStartEnd(
+            earliestPossibleSoFar, event.getStartTimeLong(), /* inclusive= */ true));
+      }
+      earliestPossibleSoFar = Math.max(earliestPossibleSoFar, event.getEndTimeLong());
+    }
+    // The end of the work hours is potentially never included so we check.
+    if (workHoursEndTimeSeconds - earliestPossibleSoFar > 0) {
```
jamesnam-google: Equivalent to,

workHoursEndTimeSeconds > earliestPossibleSoFar

portfolio/src/main/java/com/google/sps/data/TimeRange.java
```diff
@@ -0,0 +1,79 @@
...
+// limitations under the License.
+
+package com.google.sps.data;
+
+/**
+ * Class representing a span of time, enforcing properties (e.g. start comes before end) and
+ * providing methods to make ranges easier to work with (e.g. {@code overlaps}).
+ */
+public final class TimeRange {
+  private final long start;
```
rgcase: Can we use `Instant` here instead of `long`?

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
```diff
@@ -0,0 +1,116 @@
...
+      Comparator.comparing(Task::getDurationSeconds).thenComparing(Task::getName);
+
+  /**
+   * This method returns a Task Collection which the greedy algorithm
+   * determines should be scheduled for the person.
+   */
+  public Collection<Task> greedy(Collection<CalendarEvent> events, Collection<Task> tasks,
+      long workHoursStartTimeSeconds, long workHoursEndTimeSeconds) {
+    List<CalendarEvent> eventsList = new ArrayList<CalendarEvent>(events);
+    Task[] tasksArray = tasks.toArray(new Task[tasks.size()]);
```
jamesnam-google: As I wrote in my comment below, consider using list instead of array, unless there is a strong reason to use array.

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
```diff
@@ -0,0 +1,116 @@
...
+    Collections.sort(eventsList, sortAscending);
+    Arrays.sort(tasksArray, 0, tasks.size(), sortByDuration);
+    TimeRange[] availableTimes =
```
jamesnam-google: Similar comment here about list vs array.

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
```diff
@@ -0,0 +1,116 @@
...
+        emptyTimeRanges(eventsList, workHoursStartTimeSeconds, workHoursEndTimeSeconds);
+
+    List<Task> scheduledTasks = new ArrayList<Task>();
+    // Index of TimeRange we are in
+    int i = 0;
+    // Index of Task we are trying to schedule
+    int j = 0;
+    // Integer indicating the start time (in seconds) we are currently trying to
+    // schedule events in
+    long scheduleTimeSeconds = 0;
```
jamesnam-google: Consider using Instant wherever possible.

portfolio/src/main/java/com/google/sps/data/CalendarEvent.java
jamesnam-google: By the way, do we really need this method? I think the callers can choose to invoke getEpochSecond themselves. I think that is one major benefit of returning Instant; the callers can decide how to work with the instant.

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
jamesnam-google: Also does this class have any instance variable? Maybe it can be a static class, as well as all of its methods.

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
```diff
@@ -0,0 +1,116 @@
...
+
+    // This will iterate through the time ranges and tasks and if one can be
+    // scheduled then it will be (and we move onto the next task) otherwise
+    // we move on to the next range (this is because the tasks are sorted by
+    // duration so if one task did not fit in the given range then we know no
+    // later ones will fit either). We create new Task objects for the result
+    // so data structures passed in are never changed.
+    while (i < availableTimes.length && j < tasksArray.length) {
+      TimeRange timeRange = availableTimes[i];
```
jamesnam-google: Consider renaming to availableTimeRange

portfolio/src/main/java/com/google/sps/servlets/ScheduleServlet.java
```diff
@@ -0,0 +1,93 @@
...
+    String resultJSON = gson.toJson(scheduledTasks);
+    System.out.println(resultJSON);
+
+    response.setContentType("application/json");
+    response.getWriter().println(resultJSON);
+  }
+
+  private static Collection<CalendarEvent> collectEventsFromJson(JSONArray eventsArray) {
+    Collection<CalendarEvent> events = new ArrayList<CalendarEvent>();
+    for (int i = 0; i < eventsArray.length(); i++) {
```
rgcase: I think you should be able to use `for (JSONObject eventJsonObject : eventsArray) { ... }` rather than indexing into the `JSONArray`.

portfolio/src/main/java/com/google/sps/servlets/ScheduleServlet.java
```diff
@@ -0,0 +1,93 @@
...
+      JSONObject eventAsJSONObject = eventsArray.getJSONObject(i);
+      String name = (String) eventAsJSONObject.get("name");
```
rgcase: I think there are built in methods like `.getString(...)` that you can use here. Here is the API I found https://stleary.github.io/JSON-java/org/json/JSONArray.html

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
jamesnam-google: It would be simpler to use ternary operator in the constructor.

ex:
Task scheduledTask = new Task(
    task.getName(),
    task.getDescription().isPresent() ? task....get() : null,
    ...);

rgcase (commented):
Top-level comment: I think we should be able to change all the `long`s to `Instant` or `Duration`.  It'll make it much safer doing computations on these values and less likely to introduce a bug later.

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
jamesnam-google: Consider renaming to currentScheduleTimeSeconds

portfolio/src/main/java/com/google/sps/data/ScheduleRequest.java
```diff
@@ -0,0 +1,75 @@
+package com.google.sps.data;
+
+import java.time.Instant;
+import java.util.Collection;
+/**
```
jamesnam-google: Line break above this line. Also, can you run Java autoformatter and ESLINT on the code change?

portfolio/src/main/java/com/google/sps/data/ScheduleRequest.java
```diff
@@ -0,0 +1,75 @@
...
+    return workHoursStartTime;
+  }
+  public Instant getWorkHoursEndTimeInstant() {
+    return workHoursEndTime;
+  }
+
+  // These methods return the instant as a number of seconds from 00:00:00 UTC
+  // on 1 January 1970.
+
+  public long getWorkHoursStartTimeLong() {
```
jamesnam-google: As I suggested earlier, I think this class can just have methods that return Instant. Then the callers can decide which methods to call on the Instant (like getEpochSecond).

portfolio/src/main/java/com/google/sps/data/ScheduleRequest.java
```diff
@@ -0,0 +1,75 @@
...
+
+  /* Because all four fields are required, the following getters won't return null. */
+  public Collection<CalendarEvent> getEvents() {
+    return events;
+  }
+
+  public Collection<Task> getTasks() {
+    return tasks;
+  }
+  public Instant getWorkHoursStartTimeInstant() {
```
jamesnam-google: Line break between methods

portfolio/src/main/java/com/google/sps/data/ScheduleRequest.java
```diff
@@ -0,0 +1,75 @@
...
+   * @param workHoursStartTimeString: a string representation of the work hours start
+   *     time. The string is in format YYYY-MM-DDTHH:MM:SSZ
+   * @param workHoursEndTimeString: a string representation of the work hours end time.
+   *     The string's format is the same as startTimeString's.
+   * All of these fields are required for a request however, events and tasks can be.
+   * empty in the case where there are no events or no tasks. For example an empty
+   * schedule for that day or the user hits start scheduling before they add any tasks
+   * (maybe by accident).
+   */
+  public ScheduleRequest(Collection<CalendarEvent> events, Collection<Task> tasks,
```
jamesnam-google: Let's have one param per line.

portfolio/src/main/java/com/google/sps/data/Task.java
```diff
@@ -13,32 +14,55 @@
...
+    this.description = Optional.ofNullable(description);
+    this.duration = Duration.ofMinutes(durationMinutes);
+    this.priority = priority;
+  }
+
+  /**
+   * The constructor allows for Tasks to have their scheduled time be set. New tasks being made
+   * for scheduling purposes use this constructor instead.
+   */
+  public Task(String name, String description, long durationMinutes, TaskPriority priority,
```
jamesnam-google: Let's have one param per line

portfolio/src/main/java/com/google/sps/data/Task.java
```diff
@@ -53,7 +77,33 @@ public Duration getDuration() {
...
-  public Integer getPriority() {
+  public TaskPriority getPriority() {
+    return priority;
+  }
+
+  public Integer getPriorityInt() {
     return priority.getPriority();
   }
+
+  public long getDurationSeconds() {
```
jamesnam-google: Consider removing this method, because the caller can get the duration and invoke getSeconds()

portfolio/src/main/java/com/google/sps/servlets/ScheduleServlet.java
```diff
@@ -0,0 +1,93 @@
...
+    JSONArray tasksArray = jsonFromRequest.getJSONArray("tasks");
+    String workHoursStartTimeString = (String) jsonFromRequest.get("startTime");
+    String workHoursEndTimeString = (String) jsonFromRequest.get("endTime");
+
+    Collection<CalendarEvent> events = collectEventsFromJson(eventsArray);
+    Collection<Task> tasks = collectTasksFromJson(tasksArray);
+
+    ScheduleRequest scheduleRequest =
+        new ScheduleRequest(events, tasks, workHoursStartTimeString, workHoursEndTimeString);
+    FindSchedule findSchedule = new FindSchedule();
```
jamesnam-google: I think FindSchedule can be static. Then you don't have to instantiate it, and just call

FindSchedule.greedy(...)

portfolio/src/main/java/com/google/sps/servlets/ScheduleServlet.java
```diff
@@ -0,0 +1,93 @@
...
+    Collection<Task> scheduledTasks =
+        findSchedule.greedy(scheduleRequest.getEvents(), scheduleRequest.getTasks(),
```
jamesnam-google: Consider having one param per line.

portfolio/src/main/java/com/google/sps/servlets/ScheduleServlet.java
```diff
@@ -0,0 +1,93 @@
...
+            scheduleRequest.getWorkHoursStartTimeLong(), scheduleRequest.getWorkHoursEndTimeLong());
+
+    Gson gson = new Gson();
+    String resultJSON = gson.toJson(scheduledTasks);
```
jamesnam-google: Typically variable names have camel-case, including acronyms.

ex:
resultJson

portfolio/src/main/java/com/google/sps/servlets/ScheduleServlet.java
```diff
@@ -0,0 +1,93 @@
...
+    System.out.println(resultJSON);
+
+    response.setContentType("application/json");
+    response.getWriter().println(resultJSON);
+  }
+
+  private static Collection<CalendarEvent> collectEventsFromJson(JSONArray eventsArray) {
```
jamesnam-google: Consider renaming to, collectEventsFromJsonArray

portfolio/src/main/java/com/google/sps/servlets/ScheduleServlet.java
```diff
@@ -0,0 +1,93 @@
...
+      String name = (String) eventAsJSONObject.get("name");
+      String startTime = (String) eventAsJSONObject.get("startTime");
+      String endTime = (String) eventAsJSONObject.get("endTime");
+      CalendarEvent newEvent = new CalendarEvent(name, startTime, endTime);
+      events.add(newEvent);
+    }
+    return events;
+  }
+
+  private static Collection<Task> collectTasksFromJson(JSONArray tasksArray) {
```
jamesnam-google: Similar comment here about method name

portfolio/src/main/java/com/google/sps/servlets/ScheduleServlet.java
```diff
@@ -0,0 +1,93 @@
...
+    Collection<Task> tasks = new ArrayList<Task>();
+    for (int i = 0; i < tasksArray.length(); i++) {
+      JSONObject taskAsJSONArray = tasksArray.getJSONObject(i);
+      String name = (String) taskAsJSONArray.get("name");
+      String description = (String) taskAsJSONArray.get("description");
+      long duration = Long.valueOf((int) taskAsJSONArray.get("durationMinutes"));
```
jamesnam-google: duration => durationMinutes

That being said, I would encourage that you perform the conversion to Duration as soon as possible, then pass it as a param to Task constructor rather than passing minutes. Same with Instant.

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
```diff
@@ -0,0 +1,116 @@
...
+      Task task = tasksArray[j];
+      // Either time is already past the start of the time range or we should
+      // update it (maybe this is our first iteration in the range).
+      scheduleTimeSeconds = Math.max(scheduleTimeSeconds, timeRange.start());
+      // The task can be scheduled in the current time range.
+      if (scheduleTimeSeconds + task.getDurationSeconds() <= timeRange.end()) {
+        Task scheduledTask;
+        if (task.getDescription().isPresent()) {
+          scheduledTask =
+              new Task(task.getName(), task.getDescription().get(), task.getDuration().toMinutes(),
```
jamesnam-google: Consider having one param per line, here and below.

jamesnam-google (changes requested):

portfolio/src/main/java/com/google/sps/data/CalendarEvent.java
talvarez0507: I removed the methods.

talvarez0507: I removed the methods.

portfolio/src/main/java/com/google/sps/data/CalendarEvent.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
talvarez0507: I left a TODO to refactor it into an abstract class for the next PR and made it static.

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
talvarez0507: I implemented the ScheduledTask wrapper you suggested so now it returns a collection of those instead.

talvarez0507: Changed method name to be shortestTaskFirst and updated comment.

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
talvarez0507: Done.

talvarez0507: Renamed to currentScheduleTime because I refactored it into an Instant.

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
talvarez0507: Done.

talvarez0507: Refactored using a new class so this wasn't necessary.

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
talvarez0507: Changed it to return a List instead and renamed it to getEmptyTimeRanges.

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
talvarez0507: It now fits on one line because of the refactoring.

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/data/ScheduleRequest.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/data/ScheduleRequest.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/data/ScheduleRequest.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/data/ScheduleRequest.java
talvarez0507: Deleted the methods.

portfolio/src/main/java/com/google/sps/data/Task.java
talvarez0507: I implemented ScheduledTask.

portfolio/src/main/java/com/google/sps/data/Task.java
talvarez0507: I reverted back to Duration.

portfolio/src/main/java/com/google/sps/data/Task.java
talvarez0507: I changed some names and it fit on one line again.

portfolio/src/main/java/com/google/sps/data/Task.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/data/Task.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/data/TimeRange.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/data/TimeRange.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/servlets/ScheduleServlet.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/servlets/ScheduleServlet.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/servlets/ScheduleServlet.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/servlets/ScheduleServlet.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/servlets/ScheduleServlet.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/servlets/ScheduleServlet.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/servlets/ScheduleServlet.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/servlets/ScheduleServlet.java
talvarez0507: org.json.simple.JSONArray implements a raw Iterator so I changed it to a for each loop but still had to cast each Object into a JSONObject.

portfolio/src/main/java/com/google/sps/data/CalendarEvent.java
```diff
@@ -5,34 +5,32 @@
 /** Models a calendar event. It can be used for both creation and import flow. */
 public class CalendarEvent {
   private String name;
-  private final Instant startTime;
-  private final Instant endTime;
+  private final Instant startTimeInstant;
```
rgcase: I don't think you need the `*Instant` suffix anywhere in this class, the type is right there.

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
```diff
@@ -0,0 +1,118 @@
...
+    Collections.sort(tasksList, sortByTaskDurationThenName);
+    List<TimeRange> availableTimes =
+        getEmptyTimeRanges(eventsList, workHoursStartTime, workHoursEndTime);
+
+    List<ScheduledTask> scheduledTasks = new ArrayList<ScheduledTask>();
+    // Index of TimeRange we are in.
+    int i = 0;
+    // Index of Task we are trying to schedule.
+    int j = 0;
+    // Integer indicating the start time (in minutes) we are currently trying to
```
rgcase: nit: The comment still says "Integer" instead of "Instant". You can remove the "(in minutes)" after making the change too.

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
```diff
@@ -0,0 +1,118 @@
...
+    List<CalendarEvent> eventsList = new ArrayList<CalendarEvent>(events);
+    List<Task> tasksList = new ArrayList<Task>(tasks);
+    Collections.sort(eventsList, sortByEventStartTimeAscending);
+    Collections.sort(tasksList, sortByTaskDurationThenName);
+    List<TimeRange> availableTimes =
+        getEmptyTimeRanges(eventsList, workHoursStartTime, workHoursEndTime);
+
+    List<ScheduledTask> scheduledTasks = new ArrayList<ScheduledTask>();
+    // Index of TimeRange we are in.
+    int i = 0;
```
rgcase: Rather than have these comments, maybe it's more clear if you just name the variables appropriately, like `rangeIndex` and `taskIndex` or something like that. Then when you're reading the code below you don't have to look back up here all the time to find out which is which.

portfolio/src/test/java/com/google/sps/FindScheduleTest.java
```diff
@@ -0,0 +1,130 @@
...
+
+    Collection<ScheduledTask> actual = FindSchedule.shortestTaskFirst(events, tasks, TIME_0900, TIME_1700);
+    Collection<ScheduledTask> expected = Arrays.asList();
+
+    Assert.assertEquals(expected, actual);
+  }
+
+  // TODO(tomasalvarez): rewrite all these tests to reflect changes made 
+
+  /*
```
jamesnam-google: Can we have unit test that goes with the code change made? Also, please do not have commented out code here like this. It is better to remove from the code-base (you can always refer to the pull request to add it back).

portfolio/src/test/java/com/google/sps/FindScheduleTest.java
```diff
@@ -0,0 +1,130 @@
...
+  private static Instant TIME_1100 = BEGINNING_OF_DAY.plusSeconds(11 * TIME_60_MINUTES);
+  private static Instant TIME_1200 = BEGINNING_OF_DAY.plusSeconds(12 * TIME_60_MINUTES);
+  private static Instant TIME_1300 = BEGINNING_OF_DAY.plusSeconds(13 * TIME_60_MINUTES);
+  private static Instant TIME_1400 = BEGINNING_OF_DAY.plusSeconds(14* TIME_60_MINUTES);
+  private static Instant TIME_1500 = BEGINNING_OF_DAY.plusSeconds(15 * TIME_60_MINUTES);
+  private static Instant TIME_1600 = BEGINNING_OF_DAY.plusSeconds(16* TIME_60_MINUTES);
+  private static Instant TIME_1700 = BEGINNING_OF_DAY.plusSeconds(17 * TIME_60_MINUTES);
+  private static Instant TIME_1800 = BEGINNING_OF_DAY.plusSeconds(18 * TIME_60_MINUTES);
+
+
```
jamesnam-google: Unnecessary line break

portfolio/src/main/webapp/request.js
```diff
@@ -0,0 +1,25 @@
...
+  /**
+   * All fields are required for this object.
+   * @param events: events happening the day of the scheduling,
+   * @param tasks: the tasks a user would like to schedule on that day
+   * @param startTime: a string representation of the work hours start
+   *     time. The string is in format YYYY-MM-DDTHH:MM:SSZ
+   * @param endTime: a string representation of the work hours end time.
+   *     The string's format is the same as startTimeString's.
+   */
+  constructor(events, tasks, startTime, endTime) {
```
jamesnam-google: Can this method take in the Date Javascript object, rather than a String?

portfolio/src/main/java/com/google/sps/data/ScheduleRequest.java
```diff
@@ -0,0 +1,68 @@
...
+   * @param workHoursEndTimeString: a string representation of the work hours end time. The string's
+   *     format is the same as startTimeString's. All of these fields are required for a request
+   *     however, events and tasks can be. empty in the case where there are no events or no tasks.
+   *     For example an empty schedule for that day or the user hits start scheduling before they
+   *     add any tasks (maybe by accident).
+   */
+  public ScheduleRequest(
+      Collection<CalendarEvent> events,
+      Collection<Task> tasks,
+      String workHoursStartTimeString,
```
jamesnam-google: Is it possible for this method to take in Instant for start time and end time?

jamesnam-google (changes requested):

portfolio/src/test/java/com/google/sps/FindScheduleTest.java
```diff
@@ -0,0 +1,130 @@
...
+        Arrays.asList(new CalendarEvent("Event 1", "2020-06-25T00:00:00Z", "2020-06-26T00:00:00Z"));
+    Collection<Task> tasks = Arrays.asList();
+
+    Collection<ScheduledTask> actual = FindSchedule.shortestTaskFirst(events, tasks, TIME_0900, TIME_1700);
+    Collection<ScheduledTask> expected = Arrays.asList();
+
+    Assert.assertEquals(expected, actual);
+  }
+
+  // TODO(tomasalvarez): rewrite all these tests to reflect changes made
```
rgcase: Do you plan to fix these tests in this PR?

rgcase (changes requested):

portfolio/src/test/java/com/google/sps/FindScheduleTest.java
talvarez0507: I rewrote the tests after I implemented the abstract class you suggested. It feels counter-intuitive to fix this now since I already rewrote the tests based on that implementation. I will remove the file from this PR and include all of the changes in the next.

portfolio/src/test/java/com/google/sps/FindScheduleTest.java
talvarez0507: Done.

portfolio/src/main/webapp/request.js

portfolio/src/main/java/com/google/sps/data/ScheduleRequest.java
talvarez0507: Since the servlet handles all of the unpacking at this point, I just removed this file.

portfolio/src/main/java/com/google/sps/data/CalendarEvent.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
talvarez0507: Done.

portfolio/src/test/java/com/google/sps/FindScheduleTest.java
talvarez0507: I rewrote the tests after I refactored a lot of the code. I will remove the file from this PR and on the next PR the updated files will include all of the changes.

portfolio/src/main/java/com/google/sps/data/Task.java
```diff
@@ -53,7 +52,22 @@ public Duration getDuration() {
...
+
+  // These methods provide a way to tell if two objects are both Task objects
+  // and have all the same fields
+
+  @Override
+  public boolean equals(Object other) {
+    return other instanceof Task && equals(this, (Task) other);
+  }
+
+  private static boolean equals(Task a, Task b) {
```
jamesnam-google, line 67: I think you'll want to start using task ID soon. The current approach won't work well when you begin to integrate with Tasks API.

portfolio/src/main/java/com/google/sps/data/TimeRange.java
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  private static boolean equals(TimeRange a, TimeRange b) {
+    return a.start.equals(b.start) && a.duration.equals(b.duration);
+  }
+
+  /**
+   * Creates a {@code TimeRange} from {@code start} to {@code end}. Whether or not {@code end} is
+   * included in the range will depend on {@code inclusive}. If {@code inclusive} is {@code true},
+   * then @{code end} will be in the range.
```
jamesnam-google: {@code end} instead of @{code end}

portfolio/src/main/webapp/request.js
```diff
@@ -0,0 +1,23 @@
+/**
+ * Models the requests sent to the algorithm for scheduling. This class is
+ * useful when the request information needs to be sent to the scheduling
+ * servlet through a POST request. This class contains all the request
+ * information that can be easily converted into a JSON string.
+ */
+class ScheduleRequest {
+  /**
+   * All fields are required for this object.
+   * @param events: events happening the day of the scheduling,
```
jamesnam-google: Consider clarifying the type of events and tasks. Are they Javascript arrays of * ?

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
```diff
@@ -0,0 +1,112 @@
...
+      List<CalendarEvent> events, Instant startTime, Instant endTime) {
+    List<TimeRange> possibleTimes = new ArrayList<TimeRange>();
+    // This represents the earliest time that we can schedule a window for the
+    // meeting. As events are processed, this changes to their end times.
+    Instant earliestNonScheduledInstant = startTime;
+    for (CalendarEvent event : events) {
+      // Make sure that there is some time between the events and it is not
+      // later than the person's working hours ending time.
+      if (event.getStartTimeInstant().isAfter(earliestNonScheduledInstant)
+          && (!event.getStartTimeInstant().isAfter(endTime))) {
```
jamesnam-google: Unnecessary parents around !event....

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
```diff
@@ -0,0 +1,112 @@
...
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.List;
+
+// TODO(tomasalvarez): Refactor this class into an abstract class where each of
+// the scheduling algorithms can be their own class that extends this one.
+
+public class FindSchedule {
```
jamesnam-google, line 28: Consider adding TODO comment to add test coverage.

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
```diff
@@ -0,0 +1,112 @@
...
+        scheduledTasks.add(scheduledTask);
+        currentScheduleTime = currentScheduleTime.plusSeconds(task.getDuration().getSeconds());
+        taskIndex++;
+      } else {
+        rangeIndex++;
+      }
+    }
+
+  /**
+   * This method returns an TimeRange array which represent the periods of time that are empty of
```
jamesnam-google: array => list. Arrays and Lists are different concepts in Java.

portfolio/src/main/java/com/google/sps/servlets/ScheduleServlet.java
```diff
@@ -0,0 +1,92 @@
...
+        algorithm.schedule(events, tasks, workHoursStartTime, workHoursEndTime);
+
+    Gson gson = new Gson();
+    String resultJson = gson.toJson(scheduledTasks);
+
+    response.setContentType("application/json");
+    response.getWriter().println(resultJson);
+  }
+
+  private static Collection<CalendarEvent> collectEventsFromJsonArray(JSONArray eventsArray) {
```
jamesnam-google, line 61: Consider adding TODO comment to add test for this method and the one below.

portfolio/src/main/java/com/google/sps/data/TimeRange.java
```diff
@@ -0,0 +1,73 @@
...
+package com.google.sps.data;
+
+import java.time.Duration;
+import java.time.Instant;
+
+/**
+ * Class representing a span of time, enforcing properties (e.g. start comes before end) and
+ * providing methods to make ranges easier to work with (e.g. {@code overlaps}).
+ */
+public final class TimeRange {
```
jamesnam-google, line 26: Consider adding TODO comment to add test for this class.

jamesnam-google (approved):

portfolio/src/main/java/com/google/sps/data/TimeRange.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/servlets/ScheduleServlet.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/data/FindSchedule.java
talvarez0507: Done.

portfolio/src/main/webapp/request.js

portfolio/src/main/java/com/google/sps/data/TimeRange.java
talvarez0507: Done.

rgcase (approved):

PR #3, opened 2020-06-25
Front-end changes for each calendar event's event card

hollyyuqizheng:
I used the 'card' class to model each calendar event. There is also a delete button for each calendar event. 

If an event is added successfully, it will be displayed as a new card:
Screenshot 2020-06-25 at 2 42 50 PM

If an event has the same name, start and end times as one of the events already added, there will be a warning message, and the duplicate event won't get added.
Screenshot 2020-06-25 at 5 04 32 PM

If the user makes a mistake and enters an end time that is earlier than the start time, there will also be a warning message, and the duplicate event won't get added.
Screenshot 2020-06-25 at 5 05 09 PM

Also, **please ignore the files under target folder**. For some reason they are not being git-ignored this time, but those files are most likely exact copies of files in src folder.


portfolio/src/main/webapp/calendar.js
```diff
@@ -69,14 +91,65 @@ function getTimeObject(timeString) {
   return new Date(currentYear, currentMonth, currentDate, timeHour, timeMinute);
 }
 
-/** Creates a single element for a new calendar event. */
+/** Creates a card element for a new calendar event. */
```
talvarez0507, line 94: I think it should only be /* for the comment in this case.

portfolio/src/main/webapp/calendar.js
```diff
@@ -69,14 +91,65 @@ function getTimeObject(timeString) {
...
-  eventList.appendChild(newEventElement);
+  eventList.appendChild(newEventCard);
+
+  // The delete button removes the event's card from the UI. 
+  deleteButton.onclick = function(newEventCard) {
+    newEventCard.target.closest('div.card').remove(); 
+  }
+}
+
+/** Collects and returns all the events currently displayed on the UI. */
```
talvarez0507, line 134: Similarly here.

portfolio/src/main/webapp/calendar.js
```diff
@@ -69,14 +91,65 @@ function getTimeObject(timeString) {
...
+  // Each element in this set is a Json string.
+  allEventJson = new Set();
+
+  const eventList = document.getElementById('new-event-list');
+
+  // Looks at each event card and scrapes the event's name and
+  // start and end times from the HTML elements.
+  // Add all event information to a set of all Json strings.
+  eventList.childNodes.forEach((eventCard) => {
+    const eventName = eventCard.childNodes[0].childNodes[0].innerText;
```
talvarez0507: It seems like eventCard.childNodes[0] is used repeatedly, maybe we can make this its own variable to improve readability.

hollyyuqizheng: Updated.

portfolio/target/portfolio-1/calendar.js
```diff
@@ -1,86 +1,141 @@
...
+  }
 }
 
 /** Collects all the events currently displayed on the UI. */
 function collectAllEvents() {
   // A set for all calendar events displayed on the UI.
   // Each element in this set is a Json string. 
   allEventJson = new Set(); 
-
+
```
raulcruise, line 133: Remove extra space.

portfolio/target/surefire-reports/TEST-com.google.sps.CalendarEventTest.xml
```diff
@@ -61,5 +61,5 @@
     <property name="sun.io.unicode.encoding" value="UnicodeLittle"/>
     <property name="sun.cpu.isalist" value=""/>
   </properties>
-  <testcase classname="com.google.sps.CalendarEventTest" name="TestTimeStringConversion" time="0.097"/>
-</testsuite>
+  <testcase classname="com.google.sps.CalendarEventTest" name="testTimeStringConversion" time="0.11"/>
+</testsuite>
```
raulcruise, line 65: New line after file.

raulcruise (approved):

portfolio/target/portfolio-1/calendar.js
hollyyuqizheng: The file is under target, so please ignore.

talvarez0507 (approved):

portfolio/src/main/webapp/calendar.js
```diff
@@ -69,14 +103,70 @@ function getTimeObject(timeString) {
...
+
+  const eventList = document.getElementById('new-event-list');
+
+  // Looks at each event card and scrapes the event's name and
+  // start and end times from the HTML elements.
+  // Add all event information to a set of all Json strings.
+  eventList.childNodes.forEach((eventCard) => {
+    const eventCardBody = eventCard.childNodes[0];
+    const eventName = eventCardBody.childNodes[0].innerText;
+    const startTime = new Date(eventCardBody.childNodes[1].innerText);
```
jamesnam-google, line 151: Consider using jquery for date picker.
https://jqueryui.com/datepicker/

But feel free to leave a TODO for the date picker.

portfolio/src/main/webapp/calendar.js
```diff
@@ -69,14 +103,70 @@ function getTimeObject(timeString) {
...
 }
 
-/** Creates a single element for a new calendar event. */
+/** Creates a card element for a new calendar event. */
+/** 
+ * TODO(hollyyuqizheng) Look into refactoring the card creation code 
+ * into a JavaScript class. This card element will be used in different 
+ * sections on the UI, so this refactoring will make things simpler. 
+ */
 function updateCalendarEventList(newCalendarEvent) {
```
jamesnam-google, line 99: Let's just have one method comment rather than two. Creates [...] and TODO can belong to the same method comment.

/**
 * Creates...
 * TODO(...): ...

portfolio/src/main/webapp/calendar.js
```diff
@@ -36,7 +38,11 @@ function createNewCalendarEvent() {
   const endTime =
       getTimeObject(document.getElementById('new-event-end-time').value);
 
-  const eventName = document.getElementById('new-event-name').value;
+  // If user enters an empty event name, sets the name to a default string.
+  let eventName = document.getElementById('new-event-name').value;
+  if (!eventName) {
+    eventName = 'New Event';
```
jamesnam-google: Consider trying to make 'New Event' some class constant. Not sure how to do it in a Javascript class like this, but maybe you can do a bit of research to see whether it's possible.

portfolio/src/main/webapp/calendar.js
```diff
@@ -49,11 +55,39 @@ function createNewCalendarEvent() {
...
+
+    if (!doesEventExist) {
+      document.getElementById('event-duplicate-warning').style.visibility =
+          'hidden';
+      updateCalendarEventList(newCalendarEvent);
+    } else {
+      document.getElementById('event-duplicate-warning').style.visibility =
+          'visible';
+    }
     document.getElementById('new-event-name').value = 'New Event';
```
jamesnam-google: Similarly let's see if we can make 'New Event' some constant that does not need to be hard-coded in multiple places.

portfolio/src/main/webapp/calendar.js
```diff
@@ -27,7 +27,9 @@ class CalendarEvent {
  * CalendarEvent from this information. Adds the new event to the
  * event list.
  */
-function createNewCalendarEvent() {
+function createNewCalendarEventUserInput() {
+  setWarningsToHidden();
```
jamesnam-google: It wasn't clear to me why we have to call this method here. Is it to clear existing warnings? Consider adding some comment.

portfolio/src/main/webapp/style.css
```diff
@@ -1,9 +1,8 @@
-#event-end-time-warning {
-  visibility: hidden; 
+#event-end-time-warning,
+#event-duplicate-warning {
+  visibility: hidden;
```
jamesnam-google: (Optional) I think it might be simpler to rely on bootstrap class names to control default visibility. See this article for example. So these elements can have the class name "d-none".
https://getbootstrap.com/docs/4.3/utilities/display/

portfolio/src/main/webapp/calendar.js
jamesnam-google: Maybe you can have constants.js that exports some constants, for example.

portfolio/src/main/webapp/calendar.js
```diff
@@ -49,11 +55,39 @@ function createNewCalendarEvent() {
     document.getElementById('event-end-time-warning').style.visibility =
```
jamesnam-google: It can be simpler to use jquery to show/hide elements.

Sample code:
$('#event-end-time-warning').hide();
$('#event-end-time-warning').show();

That being said, using jquery is not a requirement. I see the benefits of learning vanilla javascript as well.

Reference:
https://api.jquery.com/show/

portfolio/src/main/webapp/calendar.js
```diff
@@ -69,14 +103,70 @@ function getTimeObject(timeString) {
...
-  eventList.appendChild(newEventElement);
+  eventList.appendChild(newEventCard);
+
+  // The delete button removes the event's card from the UI.
+  deleteButton.onclick = function(newEventCard) {
+    newEventCard.target.closest('div.card').remove();
+  }
+}
+
+/** Collects and returns all the events currently displayed on the UI. */
```
jamesnam-google: Can we indicate in the comment what it returns? Does it return an array of strings? JSONs?

I think it returns JSON strings, but I think it should return a list of CalendarEvent's instead. I think it should be up to the caller how they want to handle it further.

portfolio/src/main/webapp/calendar.js
jamesnam-google, line 99: (Optional) You can potentially have the template element in the DOM. It can have the appropriate class names and the nested structure. Maybe it can have the d-none class name (bootstrap) to be hidden. Then you can use jquery clone method to clone it here.
https://www.w3schools.com/jquery/html_clone.asp

portfolio/src/main/webapp/calendar.js
```diff
@@ -36,7 +38,11 @@ function createNewCalendarEvent() {
   const endTime =
       getTimeObject(document.getElementById('new-event-end-time').value);
 
-  const eventName = document.getElementById('new-event-name').value;
+  // If user enters an empty event name, sets the name to a default string.
+  let eventName = document.getElementById('new-event-name').value;
+  if (!eventName) {
```
jamesnam-google: Maybe it's simpler to do,

let eventName = document.... || 'New Event';

portfolio/src/main/webapp/index.html
```diff
@@ -89,13 +88,18 @@ <h2>Add any pre-scheduled events here</h2>
...
         <br />
-        <button type="button" class="btn btn-primary" onclick="createNewCalendarEvent()">Add this task</button>
+        <button type="button" class="btn btn-primary" onclick="createNewCalendarEventUserInput()">Add this task</button>
       </div>
     </div>
 
     <div id="new-event-list-wrapper">
       <h2>Events added:</h2>
-      <h3 id="event-end-time-warning">End time must be later than start time.</h3>
+      <div class="alert alert-secondary" role="alert" id="event-end-time-warning">
```
jamesnam-google: Usually you'd have one alert element, and you can use Javascript to control the warning message. I'd recommend that. For example, initially the alert element can have the d-none (bootstrap) class name to be hidden, and it can be empty. And when you need to show it you can use javascript to set the inner text and remove the d none class.

jamesnam-google (approved):
Approving the changes with comments.

portfolio/src/main/webapp/calendar.js
hollyyuqizheng: Updated. I actually deleted the comment because I think the refactoring might not be too efficient. Tasks and calendar events have different fields and therefore different card structure, so I think leaving the creation of cards in each object's own file is not too bad of a choice.

portfolio/src/main/webapp/calendar.js
hollyyuqizheng: Updated.

portfolio/src/main/webapp/calendar.js
hollyyuqizheng: Updated.

portfolio/src/main/webapp/calendar.js
hollyyuqizheng: Updated. I added a constant variable at the beginning of this calendar.js file.

portfolio/src/main/webapp/calendar.js
hollyyuqizheng: Updated. Also, after I changed the visibility toggling to jQuery, I don't need this setWarningsToHidden function anymore.

portfolio/src/main/webapp/style.css
hollyyuqizheng: Updated. I used the "d-none" class to handle visibility.

portfolio/src/main/webapp/calendar.js
hollyyuqizheng: Updated.

portfolio/src/main/webapp/calendar.js
hollyyuqizheng: Updated. Now this function returns an array of CalendarEvent's.

portfolio/src/main/webapp/calendar.js
hollyyuqizheng: Updated.

portfolio/src/main/webapp/index.html
hollyyuqizheng: Updated. Now there is only one warning text box.

portfolio/src/main/webapp/calendar.js
hollyyuqizheng: I see the benefit of using date picker on the UI, but for this function specifically, I was more hoping to extract relevant information from each card element on the page, so using date picker seems to not help in this regard here.

portfolio/src/main/webapp/calendar.js
```diff
@@ -36,21 +41,45 @@ function createNewCalendarEvent() {
...
+    $('#event-warning')
+
     const newCalendarEvent = new CalendarEvent(eventName, startTime, endTime);
-    updateCalendarEventList(newCalendarEvent);
-    document.getElementById('new-event-name').value = 'New Event';
+
+    // Get all the events currently displayed on the UI.
+    // Look through this set of events to ensure that duplicate events
+    // do not get added again.
+    const allEventJson = collectAllEvents();
```
rgcase: nit: Does this actually contain JSON? I think they're just instances of `CalendarEvent` aren't they?

portfolio/src/main/webapp/calendar.js
```diff
@@ -36,21 +41,45 @@ function createNewCalendarEvent() {
...
+
+    let doesEventExist = false;
+    allEventJson.forEach((existingEvent) => {
```
rgcase: You can use `.reduce(...)` ([link](https://www.w3schools.com/jsref/jsref_reduce.asp)) here, then you won't have to mutate the variable. It'll look like:
```
const doesEventExist =
  allEvents.reduce(
    (newEventExists, existingEvent) => newEventExists || eventsEqual(newCalendarEvent, existingEvent),
    /* initialValue= */ false);

File: portfolio/src/main/webapp/calendar.js [hunk 18]
```diff
@@ -69,14 +98,74 @@ function getTimeObject(timeString) {
...
+  // A set for all calendar events displayed on the UI.
+  // Each element in this set is a Json string.
+  const allEvents = [];
+
+  const eventList = document.getElementById('new-event-list');
+
+  // Looks at each event card and scrapes the event's name and
+  // start and end times from the HTML elements.
+  // Add all event information to a set of all Json strings.
+  eventList.childNodes.forEach((eventCard) => {
```
rgcase: Instead of using `forEach(...)` and mutating the `allEvents` array you can use `.map(...)` to transform all the elements of an array. You can use it like:
```
const allEvents = Array.from(eventList.childNodes).map((eventCard) => {
  ...
  return event;
});
```

portfolio/target/portfolio-1/calendar.js
```diff
@@ -1,86 +1,141 @@
...
+   */
   constructor(name, startTime, endTime) {
     this.name = name;
     this.startTime = startTime;
     this.endTime = endTime;
   }
 }
 
-/** 
+let allEvents = new Set();
```
rgcase, line 24: Can this be `const` instead of `let`?

rgcase (changes requested):

portfolio/src/main/webapp/calendar.js
hollyyuqizheng: Yes, this array now only cotnains instances of CalendarEvent's.

portfolio/target/portfolio-1/calendar.js
hollyyuqizheng: Updated.

portfolio/src/main/webapp/calendar.js
hollyyuqizheng: I see, updated.

portfolio/src/main/webapp/calendar.js
hollyyuqizheng: I see, updated.

portfolio/target/portfolio-1/calendar.js
rgcase: I still see `let`. Do you have another change to push?

rgcase (approved):

PR #4, opened 2020-06-26
Set up Mocha and add sample test

jamesnam-google:
Set up Mocha for Javascript unit testing. With this change, one can run `npm test` in the repro directory, and it will run the mocha tests.


hollyyuqizheng (approved):
Thank you for setting this up.

PR #6, opened 2020-07-01
Initial integration of Calendar API

hollyyuqizheng:
This file is my initial integration of Google Calendar API. 
On the UI, the user first sees the "Log in to Google Calendar" button. When they decide to import events from their Google Calendar, they click this button, and the Google authorization popup will appear. This authorization popup should ask the user to give read permission to their Google Calendar. 

After the user grants permission, their calendar is embedded to the UI. The log in button is disabled once the user logs in. This calendar shows the user's current Google Calendar for the week. 
Screenshot 2020-07-01 at 10 00 21 AM

At this point, the user can choose to click the "Import from Google Calendar" button, which will import all events for the next 24 hours on their calendar to the UI, added as event cards in the event column. 
Screenshot 2020-07-01 at 10 07 18 AM

I also experimented with exporting new events back to Google Calendar. So far, the "Looks good to me" button adds a hard-coded event to the user's calendar called "Testing". When the button is clicked, the user is asked to give read and write permission to their Calendar. After that, the calendar view also gets refreshed so that  the newly added events show up. 
Screenshot 2020-07-01 at 10 00 37 AM

Future steps:
- I haven't found a way to add to the calendar view first without adding to the user's actual Google Calendar, and I'm hoping there is a way to do that, depending on how synced the calendar view and the actual calendar need to be.


portfolio/src/main/webapp/calendarApi.js
```diff
@@ -0,0 +1,292 @@
...
+    $('#import-calendar-message')
+        .text(
+            'You didn\'t give permission to view your Google Calendar, ' +
+            'so your calendar events cannot be viewed or imported.');
+  }
+  $('#import-calendar-message').show();
+}
+
+/** Disconnects current user authentication. */
+function revokeAccess() {
```
talvarez0507, line 91: This does not seem to be used. Is this necessary?

portfolio/src/main/webapp/calendarApi.js
```diff
@@ -0,0 +1,292 @@
...
+      .then(function(response) {
+        var events = response.result.items;
+
+        // Show message for no imported event if result list is empty.
+        if (events.length == 0) {
+          $('#empty-calendar-import-message').removeClass('d-none');
+          $('#empty-calendar-import-message').show();
+        } else {
+          $('#empty-calendar-import-message').hide(); 
+          for (i = 0; i < events.length; i++) {
```
talvarez0507: As Ryan usually suggests maybe this should be events.forEach()

portfolio/src/main/webapp/calendarApi.js
```diff
@@ -0,0 +1,292 @@
...
+        $('#calendar-view').removeClass('d-none');
+        $('#calendar-view').attr('src', calendarViewUrl);
+        $('#calendar-view').show();
+      });
+}
+
+/**
+ * Asks the user for Write access to the API scope.
+ */
+function addWriteScope() {
```
talvarez0507, line 228: This also seems to never be used.

hollyyuqizheng: This function is actually what gets triggered by the button click. This change is updated in the HTML file, which I will push in future PRs.

portfolio/src/main/webapp/calendarApi.js
hollyyuqizheng: Updated.

portfolio/src/main/webapp/calendarApi.js
hollyyuqizheng: This code is not being used right now, but it might become useful when I handle the situation where the user leaves and comes back to the page -- maybe the authentication can be saved and the user doesn't have to re-authenticate again. I'm not entirely sure now, but I think I will keep this function for right now.

talvarez0507 (approved):

raulcruise (approved):

portfolio/src/main/webapp/calendarApi.js
```diff
@@ -0,0 +1,298 @@
+// require(['dotenv']).config({path: './.env'});
```
jamesnam-google: Can we remove this line for now?

portfolio/src/main/webapp/calendarApi.js
```diff
@@ -0,0 +1,298 @@
...
+
+// For the discovery document for Google Calendar API.
+const DISCOVERY_URL =
+    'https://www.googleapis.com/discovery/v1/apis/calendar/v3/rest';
+
+const CLIENT_ID =
+    '499747085593-hvi6n4kdrbbfvcuo1c9a9tu9oaf62cr2.apps.googleusercontent.com';
+
+// Constants for error codes during OAuth process.
+const ERROR_CODE_POPUP_CLOSED = 'popup_closed_by_user';
```
jamesnam-google: Consider using Javascript object. Maybe it can be its own file as well, like errorType.js. It can export an object like

{
  'popup_closed_by_user': 'popup_closed_by_user',
  ...
}

portfolio/src/main/webapp/calendarApi.js
```diff
@@ -0,0 +1,298 @@
...
+ * Loads the API's client and auth2 modules.
+ * Calls the initClient function after the modules load.
+ */
+function initiateCalendarAuth() {
+  gapi.load('client:auth2', initClient);
+}
+
+/** Starts authentication flow based on current user's login status. */
+function initClient() {
+  // Hide the error messages by default.
```
jamesnam-google: Is this necessary? Can we use a class name like d-none in the HTML to control the default visibility?

portfolio/src/main/webapp/calendarApi.js
```diff
@@ -0,0 +1,298 @@
...
+  // These messages will appear if authentication errors are caught.
+  hideErrorMessages();
+
+  // Initializes the gapi.client object, which app uses to make API requests.
+  // Initially, the scope is read-only to view user's Google Calendar.
+  gapi.client
+      .init({
+        'apiKey': API_KEY,
```
jamesnam-google: I think we don't need the quotes in the keys.

Reference:
https://google.github.io/styleguide/jsguide.html#features-objects-mixing-keys

Ex:
.init({
  apiKey: API_KEY,
  ...

portfolio/src/main/webapp/calendarApi.js
```diff
@@ -0,0 +1,298 @@
...
+    });
+  }
+}
+
+/**
+ * Updates import message box based on the error during authentication process
+ * for importing
+ */
+function handleImportAuthError(e) {
+  $('#import-calendar-message').removeClass('d-none');
```
jamesnam-google: Consider assigning $('...') to a variable.

var $importCalendarMessage = $('#...');

portfolio/src/main/webapp/calendarApi.js
```diff
@@ -0,0 +1,298 @@
...
+        .text(
+            'You didn\'t give permission to update your Google Calendar, ' +
+            'so your calendar events cannot be exported.');
+  }
+  $('#export-calendar-message').show();
+}
+
+/** Hides the error messages when authentication completes succesfully. */
+function hideErrorMessages() {
+  $('#import-calendar-message').hide();
```
jamesnam-google: I'd suggest that you consistently rely on the d-none class name instead of using show() and hide().

portfolio/src/main/webapp/calendarApi.js
```diff
@@ -0,0 +1,298 @@
...
+        .text(
+            'It seems like you didn\'t complete the authorization process. ' +
+            'Please click the Export button again.');
+  } else if (e.error === ERROR_CODE_ACCESS_DENIED) {
+    $('#export-calendar-message')
+        .text(
+            'You didn\'t give permission to update your Google Calendar, ' +
+            'so your calendar events cannot be exported.');
+  }
+  $('#export-calendar-message').show();
```
jamesnam-google: I'd suggest that you consistently rely on the d-none class name instead of using show() and hide().

portfolio/src/main/webapp/calendarApi.js
```diff
@@ -0,0 +1,298 @@
...
+    showCalendarView(GoogleAuth.currentUser.get());
+  });
+}
+
+/**
+ * Updates import message box based on the error during authentication process
+ * for exporting
+ */
+function handleExportAuthError(e) {
+  $('#export-calendar-message').removeClass('d-none');
```
jamesnam-google: Consider assigning $('#export-calendar-message') to a variable.

portfolio/src/main/webapp/calendarApi.js
```diff
@@ -0,0 +1,298 @@
...
+  if (e.error === ERROR_CODE_POPUP_CLOSED) {
```
jamesnam-google: Consider assigning the error message to a variable, then calling text method after the if statements.

ex:
var errorMessage;
if (e.error == ...) {
  errorMessage = ...
} else if (...) {
  errorMessage = ...
} else {
  errorMessage = ...
}
$exportCalendarMessage.text(errorMessage).show();

I think you can chain in jquery like above.

portfolio/src/main/webapp/calendarApi.js
```diff
@@ -0,0 +1,298 @@
...
+    },
+  };
+
+  addOneEventToGoogleCalendar(event);
+}
+
+/** Adds an individual event to the authorized user's Google Calendar. */
+function addOneEventToGoogleCalendar(event) {
+  const request = gapi.client.calendar.events.insert(
+      {'calendarId': 'primary', 'resource': event});
```
jamesnam-google: As mentioned earlier you don't need the single quotes in the keys.

portfolio/src/main/webapp/calendarApi.js
```diff
@@ -0,0 +1,298 @@
...
+        handleExportAuthError(error);
+      })
+}
+
+/**
+ * Adds the scheduled task items back to the user's Google Calendar.
+ * TODO(hollyyuqizheng): change the hard-coded part later when scheduled 
+ * results are returned.
+ */
+function addNewEventsToGoogleCalendar() {
```
jamesnam-google: Can we avoid having a test method like this in the master branch? You can keep track of this on your own (like local file).

portfolio/src/main/webapp/calendarApi.js
```diff
@@ -0,0 +1,298 @@
...
+/**
+ * Retrieves the logged in user's Google email and uses that email
+ * to embed the user's Google Calendar on the UI.
+ * The calendar is displayed in the user's Google Calendar's time zone.
+ */
+function showCalendarView(user) {
+  const userEmail = encodeURIComponent(user.getBasicProfile().getEmail());
+
+  // Fetches the user's Calendar's time zone. 
+  gapi.client.calendar.settings.get({'setting': 'timezone'})
```
jamesnam-google: We don't need the single quotes in the keys.

portfolio/src/main/webapp/calendarApi.js
```diff
@@ -0,0 +1,298 @@
...
+      .then((responseTimeZone) => {
+        const userCalendarTimeZone =
+            encodeURIComponent(responseTimeZone.result.value);
+        const calendarViewUrl =
+            'https://calendar.google.com/calendar/embed?src=' + userEmail +
+            '&ctz=' + userCalendarTimeZone + '&mode=WEEK';
+
+        // Updates the UI so that calendar view appears.
+        $('#calendar-view').removeClass('d-none');
```
jamesnam-google: We can simplify as below.

var $calendarView = $('#...');
$calendarView.removeClass(...)
    .attr(...)
    .show();

That said, I think we should just add/remove d-none and avoid calling show() and hide().

portfolio/src/main/webapp/calendarApi.js
```diff
@@ -0,0 +1,298 @@
...
+  const timeRangeEnd = new Date();
+  timeRangeEnd.setFullYear(timeRangeStart.getFullYear());
+  timeRangeEnd.setMonth(timeRangeStart.getMonth());
+  timeRangeEnd.setDate(timeRangeStart.getDate() + 1);
+
+  // Retrieves events on the user's calendar for the day
+  // that the user has picked in the nav bar. 
+  gapi.client.calendar.events
+      .list({
+        'calendarId': 'primary',
```
jamesnam-google: We don't need the single quotes in the keys.

portfolio/src/main/webapp/calendarApi.js
```diff
@@ -0,0 +1,298 @@
...
+        'singleEvents': true,
+        'orderBy': 'startTime'
+      })
+      .then(function(response) {
+        var events = response.result.items;
+
+        // Show message for no imported event if result list is empty.
+        if (events.length == 0) {
+          const pickedDate = $('#date-picker').val();
+          $('#empty-calendar-import-message').removeClass('d-none');
```
jamesnam-google: We can assign $('#empty-calendar-import-message') to a variable, and chain.

var $message = $('#...');
$message.removeClass(...)
    .text(...)

Also, try to avoid mixing d-none and show()/hide(). I'd just add and remove class d-none.

portfolio/src/main/webapp/calendarApi.js
```diff
@@ -0,0 +1,298 @@
...
+  GoogleAuth.disconnect();
+}
+
+/** Updates the calendar view and button visibility based on login status. */
+function updateCalendarView() {
+  var user = GoogleAuth.currentUser.get();
+  var isAuthorized = user.hasGrantedScopes(SCOPE_READ_ONLY);
+  if (isAuthorized) {
+    showCalendarView(user);
+    $('#calendar-auth-button').attr('disabled', 'true');
```
jamesnam-google, line 101: See my comment below on assigning to jquery variables, and not calling show()/hide() and just adding/removing class d-none.

jamesnam-google (changes requested):

portfolio/src/main/webapp/calendarApi.js
hollyyuqizheng: Updated.

portfolio/src/main/webapp/calendarApi.js
hollyyuqizheng: Updated.

portfolio/src/main/webapp/calendarApi.js
hollyyuqizheng: Updated.

portfolio/src/main/webapp/calendarApi.js
hollyyuqizheng: Updated.

portfolio/src/main/webapp/calendarApi.js
hollyyuqizheng: Updated.

portfolio/src/main/webapp/calendarApi.js
hollyyuqizheng: I moved the lines under original hideErrorMessages function to the function that initiates authentication flow, and I use the d-none class to set their visibility to hidden by default.

portfolio/src/main/webapp/calendarApi.js
hollyyuqizheng: Updated.

portfolio/src/main/webapp/calendarApi.js
hollyyuqizheng: Updated.

portfolio/src/main/webapp/calendarApi.js
hollyyuqizheng: Updated. I only use d-none to toggle visibility now.

portfolio/src/main/webapp/calendarApi.js
hollyyuqizheng: Updated.

portfolio/src/main/webapp/calendarApi.js
hollyyuqizheng: Updated.

portfolio/src/main/webapp/calendarApi.js
hollyyuqizheng: Updated.

portfolio/src/main/webapp/calendarApi.js
hollyyuqizheng: Updated.

portfolio/src/main/webapp/calendarApi.js
hollyyuqizheng: Updated. I removed the content of the event from this function and kept only the function signature.

portfolio/src/main/webapp/calendarApi.js
hollyyuqizheng: I moved the constants into an Object in this file. I am having issues with the "require" keyword, similar to the one I have when I try to store API keys in another variable, so for right now I left the error code object in the same file.

portfolio/src/main/webapp/calendarApi.js
```diff
@@ -0,0 +1,283 @@
...
+        'Please click the Export button again.';
+  } else if (e.error === ERROR_CODES.access_denied) {
+    errorMessage =
+        'You didn\'t give permission to update your Google Calendar, ' +
+        'so your task schedule cannot be exported.'
+  } else {
+    errorMessage = 'An error occurred';
+  }
+
+  $exportCalendarMessage.removeClass('d-none').text(errorMessage);
```
jamesnam-google: nit: I think it might make more sense to call text(...) and then call removeClass(...)

portfolio/src/main/webapp/calendarApi.js
```diff
@@ -0,0 +1,283 @@
...
+        $('#calendar-view').removeClass('d-none').attr('src', calendarViewUrl);
+      });
+}
+
+/**
+ * Onclick function for 'Looks good to me, export" button.
+ * Asks the user for Write access to the API scope.
+ */
+function addWriteScope() {
+  var GoogleUser = GoogleAuth.currentUser.get();
```
jamesnam-google: Can we follow camelCase instead of CamelCase?

ex:
var googleUser = ...

portfolio/src/main/webapp/calendarApi.js
```diff
@@ -0,0 +1,283 @@
...
+  gapi.client.calendar.settings.get({setting: 'timezone'})
+      .then((responseTimeZone) => {
+        const userCalendarTimeZone =
+            encodeURIComponent(responseTimeZone.result.value);
+        const calendarViewUrl =
+            'https://calendar.google.com/calendar/embed?src=' + userEmail +
+            '&ctz=' + userCalendarTimeZone + '&mode=WEEK';
+
+        // Updates the UI so that calendar view appears.
+        $('#calendar-view').removeClass('d-none').attr('src', calendarViewUrl);
```
jamesnam-google: I think it might make more sense to call attr before removeClass

portfolio/src/main/webapp/calendarApi.js
```diff
@@ -0,0 +1,283 @@
...
+        calendarId: 'primary',
+        timeMin: timeRangeStart.toISOString(),
+        timeMax: timeRangeEnd.toISOString(),
+        showDeleted: false,
+        singleEvents: true,
+        orderBy: 'startTime'
+      })
+      .then(function(response) {
+        var events = response.result.items;
+        var emptyCalendarMessage = $('#empty-calendar-import-message');
```
jamesnam-google: Let's have the variable start with $ to denote that it is a jquery object.

portfolio/src/main/webapp/calendarApi.js
```diff
@@ -0,0 +1,283 @@
...
+
+        // Show message for no imported event if result list is empty.
+        if (events.length == 0) {
+          const pickedDate = $('#date-picker').val();
+          emptyCalendarMessage.removeClass('d-none');
```
jamesnam-google: I think it might make more sense to call text before removeClass. And then can be chained.

$emptyCalendarMessage.text(...)
    .removeClass(...);

portfolio/src/main/webapp/calendarApi.js
```diff
@@ -0,0 +1,283 @@
...
+
+// For the discovery document for Google Calendar API.
+const DISCOVERY_URL =
+    'https://www.googleapis.com/discovery/v1/apis/calendar/v3/rest';
+
+const CLIENT_ID =
+    '499747085593-hvi6n4kdrbbfvcuo1c9a9tu9oaf62cr2.apps.googleusercontent.com';
+
+// Object for all error codes
+const ERROR_CODES =
```
jamesnam-google: nit: I think the line break before { is not necessary.

ex:
const ERROR_CODES = {
  ...
};

Note the semicolon at the end.

portfolio/src/main/webapp/calendarApi.js
```diff
@@ -0,0 +1,283 @@
...
+        'It seems like you didn\'t complete the authorization process. ' +
+        'Please click the Login button again.'
+  } else if (e.error === ERROR_CODES.access_denied) {
+    errorMessage =
+        'You didn\'t give permission to view your Google Calendar, ' +
+        'so your calendar events cannot be viewed or imported.'
+  } else {
+    errorMessage = 'An error occurred.';
+  }
+  $importCalendarMessage.removeClass('d-none').text(errorMessage);
```
jamesnam-google: Consider calling text(...) before calling removeClass(...)

portfolio/src/main/webapp/calendarApi.js
```diff
@@ -0,0 +1,283 @@
...
+}
+
+/** Disconnects current user authentication. */
+function revokeAccess() {
+  GoogleAuth.disconnect();
+}
+
+/** Updates the calendar view and button visibility based on login status. */
+function updateCalendarView() {
+  var user = GoogleAuth.currentUser.get();
```
jamesnam-google: nit: user => googleUser

jamesnam-google (approved):
Approving the changes with comments

portfolio/src/main/webapp/calendarApi.js
rgcase: Also nit: I think you can use `const` instead of `var`. There are a bunch of other places throughout this file that I think you could use `const` as well.

rgcase (approved):

portfolio/src/main/webapp/calendarApi.js
hollyyuqizheng: Updated.

portfolio/src/main/webapp/calendarApi.js
hollyyuqizheng: Updated.

portfolio/src/main/webapp/calendarApi.js
hollyyuqizheng: Updated.

portfolio/src/main/webapp/calendarApi.js
hollyyuqizheng: Updated.

portfolio/src/main/webapp/calendarApi.js
hollyyuqizheng: Updated.

portfolio/src/main/webapp/calendarApi.js
hollyyuqizheng: Updated.

portfolio/src/main/webapp/calendarApi.js
hollyyuqizheng: Updated.

portfolio/src/main/webapp/calendarApi.js
hollyyuqizheng: Updated.

PR #8, opened 2020-07-08
CalendarGroup class and tests

hollyyuqizheng:
The CalendarGroup class models a collection of calendar events. These events could be manually created by the user or imported from Google Calendar. This class contains methods for calculating free time ranges given the events and start and end times of possible scheduling block, as well as methods for deleting and modifying free time ranges, which is useful in scheduling algorithms that need to update a collection of free time ranges. 

I also deleted the "inclusive" tag for TimeRange class when creating a new time range, because from our implementation so far, this tag has always been true, which means the end time of a time range is always included.


portfolio/src/test/java/com/google/sps/CalendarGroupTest.java
```diff
@@ -0,0 +1,317 @@
...
+    Assert.assertTrue(freeTimeRangesAfterDelete.get(1).equals(freeThree));
+
+    // Creates a new time range that is 10 seconds shorter than free time One.
+    Instant freeOneEndNew = freeOneEnd.minusSeconds(10);
+    TimeRange freeOneNew = TimeRange.fromStartEnd(freeOneStart, freeOneEndNew);
+
+    List<TimeRange> freeTimeRangesAfterModify =
+        eventGroup.modifyFreeTimeRange(
+            /* originalFreeTimeRange= */ freeOne, /* newFreeTimeRange= */ freeOneNew);
+    Assert.assertTrue(freeTimeRangesAfterModify.get(0).end().equals(freeOneNew.end()));
```
talvarez0507: It seems like you can also use the collection comparison here if you'd like. Alternatively, we can probably use `Assert.assertEquals(freeOneNew.end(), freeTimeRangesAfterModify.get(0).end())`

portfolio/src/test/java/com/google/sps/CalendarGroupTest.java
```diff
@@ -0,0 +1,317 @@
...
+    Instant freeTwoEnd = eventTwoStart;
+    Instant freeThreeStart = eventTwoEnd;
+    Instant freeThreeEnd = endTime;
+
+    TimeRange freeOne = TimeRange.fromStartEnd(freeOneStart, freeOneEnd);
+    TimeRange freeTwo = TimeRange.fromStartEnd(freeTwoStart, freeTwoEnd);
+    TimeRange freeThree = TimeRange.fromStartEnd(freeThreeStart, freeThreeEnd);
+
+    List<TimeRange> freeTimeRangesAfterDelete = eventGroup.deleteFreeTimeRange(freeTwo);
+    Assert.assertTrue(freeTimeRangesAfterDelete.get(0).equals(freeOne));
```
talvarez0507: Similar to the last comment, can we change this to check for the entire collection equality.

portfolio/src/test/java/com/google/sps/CalendarGroupTest.java
```diff
@@ -0,0 +1,317 @@
...
+
+    List<CalendarEvent> events = new ArrayList<CalendarEvent>();
+    Instant startTime = Instant.now();
+    Instant endTime = startTime.plusSeconds(10000);
+    TimeRange targetFreeTime = TimeRange.fromStartEnd(startTime, endTime);
+
+    CalendarGroup eventGroup = new CalendarGroup(events, startTime, endTime);
+    List<TimeRange> freeTimeRanges = eventGroup.getFreeTimeRanges();
+
+    Assert.assertTrue(freeTimeRanges.get(0).equals(targetFreeTime));
```
talvarez0507: TimeRange implements an equals, so can we compare the Lists instead? Maybe we can use an expected and actual List. For example:
`List<TimeRange> expectedFreeTimeRanges = Arrays.asList(targetFreeTime);`
`List<TimeRange> actualFreeTimeRanges = eventGroup.getFreeTimeRanges();`
`Assert.assertEquals(expectedFreeTimeRanges, actualFreeTimeRanges);`

talvarez0507 (changes requested):
The tests seem good, but the equality checking can probably be made a little simpler.

portfolio/src/test/java/com/google/sps/CalendarGroupTest.java
hollyyuqizheng: I see, updated!

portfolio/src/test/java/com/google/sps/CalendarGroupTest.java
hollyyuqizheng: Updated.

portfolio/src/test/java/com/google/sps/CalendarGroupTest.java
hollyyuqizheng: Updated.

talvarez0507 (approved):
I left some comments with some suggestions. Only the else, continue is a necessary change.

portfolio/src/main/java/com/google/sps/data/CalendarGroup.java
```diff
@@ -0,0 +1,132 @@
...
+  // the scheduling process.
+  public final Instant startTime;
+  public final Instant endTime;
+
+  // Comparator for sorting events by start time
+  private static final Comparator<CalendarEvent> sortByEventStartTimeAscending =
+      Comparator.comparing(CalendarEvent::getStartTime);
+
+  // Comparator for sorting free time ranges by start time
+  private static final Comparator<TimeRange> sortByTimeRangeStartTimeAscending =
```
talvarez0507: Maybe we can move this into the TimeRange class.

portfolio/src/main/java/com/google/sps/data/CalendarGroup.java
```diff
@@ -0,0 +1,132 @@
...
+    // meeting. As events are processed, this changes to their end times.
+    Instant earliestNonScheduledInstant = startTime;
+    for (CalendarEvent event : events) {
+      // Make sure that there is some time between the events and it is not
+      // later than the person's scheduling hours' ending time.
+      if (event.getStartTime().isAfter(earliestNonScheduledInstant)
+          && !event.getStartTime().isAfter(endTime)) {
+        possibleTimes.add(
+            TimeRange.fromStartEnd(earliestNonScheduledInstant, event.getStartTime()));
+      }
```
talvarez0507, line 77: Make sure we include the code that's missing here and a comment: 
`// This is so events outside the time period are not processed.`
`else { continue;}`

portfolio/src/main/java/com/google/sps/data/CalendarGroup.java
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  /**
+   * Deletes a time range from the list of all free times.
+   *
+   * @return the modifed list of free time ranges.
+   */
+  public List<TimeRange> deleteFreeTimeRange(TimeRange timeRange) {
+    boolean exist = false;
+    for (TimeRange freeTimeRange : allFreeTimeRanges) {
```
talvarez0507: Collections include a .contains() method. I think this entire method can just be:
``` 
if (allFreeTimeRanges.contains(timeRange)) {    
 allFreeTimeRanges.remove(timeRange);    
} else {   
  throw new IllegalArgumentException("This time range does not represent any free time range.");
return allFreeTimeRanges; 
```
Also it seems like the modification methods should not be returning anything, maybe they should be void.

portfolio/src/test/java/com/google/sps/CalendarGroupTest.java
```diff
@@ -0,0 +1,318 @@
...
+
+    Instant wrongStart = eventOneStart.minusSeconds(50);
+    Instant wrongEnd = eventOneStart;
+
+    eventGroup.deleteFreeTimeRange(TimeRange.fromStartEnd(wrongStart, wrongEnd));
+  }
+
+  /** Tests for correct free times with non-overlapping events. */
+  @Test
+  public void nonoverlappingEvents() {
```
talvarez0507: This tests the calculateFreeTimeRanges() method more specifically so maybe the name should reflect that, possibly "freeTimeRangesNonOverlappingEvents()".

portfolio/src/test/java/com/google/sps/CalendarGroupTest.java
```diff
@@ -0,0 +1,318 @@
...
+
+    CalendarEvent eventOne = new CalendarEvent("event one", eventOneStart, eventOneEnd);
+    CalendarEvent eventTwo = new CalendarEvent("event two", eventTwoStart, eventTwoEnd);
+
+    events.add(eventOne);
+    events.add(eventTwo);
+
+    CalendarGroup eventGroup = new CalendarGroup(events, startTime, endTime);
+
+    Instant freeOneStart = startTime;
```
talvarez0507: "freeOne" and similar variable names make sense in context but maybe something like "freeTimeRangeOne", "openTimeRangeOne" or "unscheduledTimeRangeOne" would be more descriptive.

portfolio/src/main/java/com/google/sps/data/CalendarGroup.java
hollyyuqizheng: Updated.

portfolio/src/main/java/com/google/sps/data/CalendarGroup.java
hollyyuqizheng: On second thought, I don't think we should include this continue statement. For example, if the earliestNonScheduledInstant is at 9:00, and we have an event from 8:30 to 9:30, we do not add any new time range into possibleTimes, but we need to shift the earliestNonScheduledInstant to 9:30 now, which is handled in the if-statement that follows, so I don't think we can skip this if statement?

portfolio/src/main/java/com/google/sps/data/CalendarGroup.java
hollyyuqizheng: I see about the contains method, I didn't realize the equals method is overriding the default one. 
In terms of the modification method, I was thinking that, since the deleting method returns the new collection, I wanted to keep both methods similar, so both would return the newly modified collection.

portfolio/src/test/java/com/google/sps/CalendarGroupTest.java
hollyyuqizheng: Updated.

portfolio/src/test/java/com/google/sps/CalendarGroupTest.java
hollyyuqizheng: Updated.

raulcruise (approved):

portfolio/src/main/java/com/google/sps/data/CalendarGroup.java
```diff
@@ -0,0 +1,126 @@
+package com.google.sps.data;
+
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.List;
+
+/** This class models a group of calendar events. */
+public class CalendarGroup {
```
jamesnam-google: Based on the comment, consider renaming to CalendarEventsGroup.

portfolio/src/main/java/com/google/sps/data/CalendarGroup.java
```diff
@@ -0,0 +1,126 @@
...
+  public List<CalendarEvent> events;
+  public List<TimeRange> allFreeTimeRanges;
+
+  // These two variables represent the start and end of
+  // possible schedule time. When these times are passed in
+  // to instantiate a CalendarGroup class, these times are
```
jamesnam-google: CalendarGroup => CalendarEventsGroup. Or you can just say "instantiate this class"

portfolio/src/main/java/com/google/sps/data/CalendarGroup.java
```diff
@@ -0,0 +1,126 @@
...
+   * @param events: a list of CalendarEvents' that represents the events already scheduled for the
+   *     user.
+   * @param startTime: start of possible scheduling blocks, of type Instant. This is the earlier of
+   *     working hour start time and the current timestamp.
+   * @param endTime: end of possible scheduling blocks, of type Instant.
+   */
+  public CalendarGroup(
+      List<CalendarEvent> events, Instant overallStartTime, Instant overallEndTime) {
+    if (events == null) {
+      throw new IllegalArgumentException("Calendar Group needs a collection of events");
```
jamesnam-google: I think it might be simpler to make the exception message look like, "Events cannot be null". Here and below.

portfolio/src/main/java/com/google/sps/data/CalendarGroup.java
```diff
@@ -0,0 +1,126 @@
...
+
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.List;
+
+/** This class models a group of calendar events. */
+public class CalendarGroup {
+  public List<CalendarEvent> events;
```
jamesnam-google: Consider clarifying in the comment that this variable will always be sorted. And I think it'd be a good idea to ensure that's the case.

portfolio/src/main/java/com/google/sps/data/CalendarGroup.java
```diff
@@ -0,0 +1,126 @@
...
+  public List<TimeRange> allFreeTimeRanges;
```
jamesnam-google: Is this variable sorted? Consider adding comment and ensuring that it is.

That being said, I'm not sure whether you need this variable, because it can be computed based on events and overall start/end times. So Maybe remove this variable, and just expose a method that returns all free time ranges.

portfolio/src/main/java/com/google/sps/data/CalendarGroup.java
```diff
@@ -0,0 +1,126 @@
...
+  public List<TimeRange> getFreeTimeRanges() {
+    return allFreeTimeRanges;
+  }
+
+  /**
+   * Deletes a time range from the list of all free times.
+   *
+   * @return the modifed list of free time ranges.
+   */
+  public List<TimeRange> deleteFreeTimeRange(TimeRange timeRange) {
```
jamesnam-google: This method and the one below seem a bit weird to me, because I think free time ranges are computed based on events and overall start/end instants. I think it would make more sense to expose methods for adding/removing calendar events. And then at any point the caller can get the list of free time ranges. So consider removing these time range modification methods. And I don't think they are called anywhere except for in tests.

portfolio/src/test/java/com/google/sps/CalendarGroupTest.java
```diff
@@ -0,0 +1,328 @@
...
+public final class CalendarGroupTest {
+
+  /** Start time of the scheduling block must not be later then end time. */
+  @Test(expected = IllegalArgumentException.class)
+  public void startTimeAfterEndTime() {
+    Instant startTime = Instant.now();
+    Instant endTimeBefore = startTime.minusSeconds(1000);
+    Instant endTimeAfter = startTime.plusSeconds(1000);
+
+    List<CalendarEvent> events = new ArrayList<CalendarEvent>();
```
jamesnam-google: We can simplify these lines with,

List<CalendarEvent> events = ImmutableList.of(
    new CalendarEvent(...));

jamesnam-google (changes requested):
Please see my comments.

portfolio/src/main/java/com/google/sps/data/CalendarGroup.java
hollyyuqizheng: Updated.

portfolio/src/main/java/com/google/sps/data/CalendarGroup.java
hollyyuqizheng: Updated.

portfolio/src/main/java/com/google/sps/data/CalendarGroup.java
hollyyuqizheng: Updated.

portfolio/src/main/java/com/google/sps/data/CalendarGroup.java
hollyyuqizheng: Updated.

portfolio/src/main/java/com/google/sps/data/CalendarGroup.java
hollyyuqizheng: I see. I was thinking of handling more of the free time range in this class, but I can imagine that all the free time range can be handled in a different class or in the same class as the scheduling algorithm? For right now, I removed these two methods and added a method for adding individual calendar event to the group, and I don't think a method for removing events will be helpful in this case.

portfolio/src/test/java/com/google/sps/CalendarGroupTest.java
hollyyuqizheng: Updated. I ended up using Arrays.asList for some of the later test cases, because the ImmutableLists's don't seem to be sorted easily.

portfolio/src/main/java/com/google/sps/data/CalendarGroup.java
jamesnam-google: Okay sounds good

portfolio/src/main/java/com/google/sps/data/CalendarEventsGroup.java
```diff
@@ -0,0 +1,102 @@
...
+    // This represents the earliest time that we can schedule a window for the
+    // meeting. As events are processed, this changes to their end times.
+    Instant earliestNonScheduledInstant = overallStartTime;
+
+    for (CalendarEvent event : events) {
+      // Make sure that there is some time between the events and it is not
+      // later than the person's scheduling hours' ending time.
+      if (event.getStartTime().isAfter(earliestNonScheduledInstant)) {
+        if (event.getEndTime().isAfter(overallEndTime)) {
+          possibleTimes.add(TimeRange.fromStartEnd(earliestNonScheduledInstant, overallEndTime));
```
jamesnam-google, line 75: This line doesn't seem correct to me. For example, suppose:
earliestNonScheduledInstant is t1
event.getStartTime() is t2
overallEndTime is t3
event.getEndTime() is t4

Here you're considering (t1, t3) as free. But it isn't, because the event starts at t2.

jamesnam-google (changes requested):

portfolio/src/main/java/com/google/sps/data/CalendarEventsGroup.java
hollyyuqizheng: I see. I updated the if-statement and added new tests.

portfolio/src/main/java/com/google/sps/data/CalendarEventsGroup.java
```diff
@@ -0,0 +1,103 @@
...
+    this.overallEndTime = overallEndTime;
+  }
+
+  /**
+   * Calculates a TimeRange list which represents the periods of time that are empty of events and
+   * lie completely inside the possible scheduling hours. Originally written by tomasalvarez,
+   * modified by hollyyuqizheng.
+   */
+  public List<TimeRange> getFreeTimeRanges() {
+    Collections.sort(events, sortByEventStartTimeAscending);
```
jamesnam-google: Consider sorting the list in other places (like constructor) so you don't have to sort it here. In other words, you can sort when you set/update the list. When you do a get operation like this one, you don't have to sort.

portfolio/src/main/java/com/google/sps/data/CalendarEventsGroup.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    return possibleTimes;
+  }
+
+  /**
+   * Adds a new event into the event group. Returns the new collection of events, sorted based on
+   * start time.
+   */
+  public List<CalendarEvent> addNewEvent(CalendarEvent event) {
```
jamesnam-google: Consider making it return void. I think the caller probably only needs the free time ranges as output as far as this class is concerned.

jamesnam-google (approved):
Approving the change with comments.

portfolio/src/main/java/com/google/sps/data/CalendarEventsGroup.java
hollyyuqizheng: I see. Updated.

portfolio/src/main/java/com/google/sps/data/CalendarEventsGroup.java
hollyyuqizheng: Updated.

portfolio/src/main/java/com/google/sps/data/CalendarEventsGroup.java
```diff
@@ -0,0 +1,102 @@
...
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.List;
+
+/** This class models a group of calendar events. */
+public class CalendarEventsGroup {
+
+  // This variable that represents a group of calendar events will
+  // always be sorted in ascending order based on their start time.
+  public List<CalendarEvent> events;
```
rgcase: nit: I think we can make this final too? It seems like we only initialize it in the constructor.

rgcase (approved):

PR #9, opened 2020-07-08
Abstract class for scheduling and ShortestTaskFirst

talvarez0507:
This PR changes FindSchedule into an abstract BaseTaskScheduler class which future algorithms will extend. It also updates the servlet to handle and use the algorithm type based on the request from the client. I also included my scheduling algorithm ShortestTaskFirst and its JUnit tests.


portfolio/src/test/java/com/google/sps/ShortestTaskFirstTest.java
```diff
@@ -0,0 +1,171 @@
...
+    ScheduledTask scheduledTask1 = new ScheduledTask(task1, TIME_0900);
+
+    Collection<ScheduledTask> actual = algorithm.schedule(events, tasks, TIME_0900, TIME_1200);
+    Collection<ScheduledTask> expected = Arrays.asList(scheduledTask1);
+    Assert.assertEquals(expected, actual);
+  }
+
+  // Makes sure that no errors are thrown if working hours is longer than a
+  // 24 hour period.
+  @Test
```
hollyyuqizheng: I was thinking that, for all the time validity checking, eg. is the user trying to schedule for a date that is before "today", is the user's working hours correct, will all happen on the client side, so ideally all the time input to any Java methods are all valid. So this test might end up not being too necessary, because the user's working hours will not be over 3 days, per checked on the client side already.

portfolio/src/test/java/com/google/sps/ShortestTaskFirstTest.java
```diff
@@ -0,0 +1,171 @@
...
+    ScheduledTask scheduledTask1 = new ScheduledTask(task1, TIME_0900);
+    ScheduledTask scheduledTask2 = new ScheduledTask(task2, TIME_1000);
+
+    Collection<ScheduledTask> actual = algorithm.schedule(events, tasks, TIME_0900, TIME_1700);
+    Collection<ScheduledTask> expected = Arrays.asList(scheduledTask1, scheduledTask2);
+    Assert.assertEquals(expected, actual);
+  }
+
+  // Makes sure that no errors are thrown for events outside working hours.
+  @Test
```
hollyyuqizheng: Maybe we could add another test for some events that lie partially outside the working hours. For example, for a 9-5 working hour, an event could be from 8:30 to 9:30, and the test would make sure that the first scheduled task is not at 9 or something.

hollyyuqizheng (changes requested):
Some suggestions on additional tests.

portfolio/src/test/java/com/google/sps/ShortestTaskFirstTest.java
talvarez0507: The client side code can always be changed by the user so this an unlikely but possible case. This test was just to ensure no errors are thrown but maybe we should add a check somewhere in CalendarGroup?

portfolio/src/test/java/com/google/sps/ShortestTaskFirstTest.java
talvarez0507: Done. Broke this test into two new ones, events partially/totally outside working hours.

portfolio/src/main/java/com/google/sps/data/ShortestTaskFirst.java
```diff
@@ -36,7 +32,7 @@
    * This method schedules tasks from shortest to longest and returns a ScheduledTask Collection
    * based on the tasks that were able to be scheduled.
    */
-  public static Collection<ScheduledTask> shortestTaskFirst(
+  public Collection<ScheduledTask> schedule(
```
hollyyuqizheng: Just a reminder for us that calling the getEmptyTimeRanges will look a bit different since it's in another class now

portfolio/src/test/java/com/google/sps/ShortestTaskFirstTest.java
```diff
@@ -0,0 +1,187 @@
...
+  private static Instant TIME_1300 = TIME_1200.plus(TIME_60_MINUTES);
+  private static Instant TIME_1400 = TIME_1300.plus(TIME_60_MINUTES);
+  private static Instant TIME_1500 = TIME_1400.plus(TIME_60_MINUTES);
+  private static Instant TIME_1600 = TIME_1500.plus(TIME_60_MINUTES);
+  private static Instant TIME_1700 = TIME_1600.plus(TIME_60_MINUTES);
+  private static Instant TIME_1800 = TIME_1700.plus(TIME_60_MINUTES);
+
+  private final TaskPriority LOW_PRIORITY = new TaskPriority(1);
+
+  // Makes sure we return an empty list in the case where no tasks are passed.
```
raulcruise: Genuinely wondering, are we supposed to use Javadoc style for unit tests as well?

portfolio/src/test/java/com/google/sps/ShortestTaskFirstTest.java
hollyyuqizheng: I see. I think this check will happen on the client side instead of in the calendar event group class, eg. I have already added some validity checking on the UI side of things, so I think the input to the scheduling algorithm should be in good shape.

hollyyuqizheng (approved):

raulcruise (approved):

portfolio/src/test/java/com/google/sps/ShortestTaskFirstTest.java
talvarez0507: Good question, I'll ask in our next pod sync.

portfolio/src/main/java/com/google/sps/data/ShortestTaskFirst.java
talvarez0507: I fixed it now rather than later to prevent us from forgetting.

portfolio/src/main/java/com/google/sps/data/BaseTaskScheduler.java
```diff
@@ -0,0 +1,32 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.data;
+
+import java.time.Instant;
+import java.util.Collection;
+
+// TODO: Add test coverage for scheduling algorithms
```
jamesnam-google: Can we have username for the TODO.

portfolio/src/main/java/com/google/sps/data/BaseTaskScheduler.java
```diff
@@ -0,0 +1,32 @@
...
+public abstract class BaseTaskScheduler {
+  public enum SchedulingAlgorithmType {
```
jamesnam-google: Consider making this enum its own file, so it can be referenced easily in the code-base.

portfolio/src/main/java/com/google/sps/data/BaseTaskScheduler.java
```diff
@@ -0,0 +1,32 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.data;
+
+import java.time.Instant;
+import java.util.Collection;
+
+// TODO: Add test coverage for scheduling algorithms
+public abstract class BaseTaskScheduler {
```
jamesnam-google: I think this class can be package-private (i.e., take out the public keyword).

portfolio/src/main/java/com/google/sps/data/BaseTaskScheduler.java
```diff
@@ -0,0 +1,32 @@
...
+  public enum SchedulingAlgorithmType {
+    SHORTEST_TASK_FIRST
+  }
+
+  /** This is the abstract method that all algorithm classes should implement. */
+  public abstract Collection<ScheduledTask> schedule(
```
jamesnam-google: I think it can be package-private as well.

portfolio/src/main/java/com/google/sps/data/BaseTaskScheduler.java
jamesnam-google: I think there should be an abstract method that returns the algorithm type.

abstract SchedulingAlgorithmType getSchedulingAlgorithmType();

And classes that implement the abstract class can override that method.

portfolio/src/main/java/com/google/sps/data/ScheduledTask.java
```diff
@@ -24,4 +28,13 @@ public Task getTask() {
   public Instant getStartTime() {
     return startTime;
   }
+
+  @Override
+  public boolean equals(Object other) {
+    return other instanceof ScheduledTask && equals(this, (ScheduledTask) other);
+  }
+
+  public static boolean equals(ScheduledTask a, ScheduledTask b) {
```
jamesnam-google, line 37: Is this method necessary? Each time we call ScheduledTask.equals(a, b), we can just call a.equals(b).

portfolio/src/main/java/com/google/sps/data/ShortestTaskFirst.java
```diff
@@ -36,17 +32,14 @@
    * This method schedules tasks from shortest to longest and returns a ScheduledTask Collection
    * based on the tasks that were able to be scheduled.
    */
-  public static Collection<ScheduledTask> shortestTaskFirst(
-      Collection<CalendarEvent> events,
-      Collection<Task> tasks,
-      Instant workHoursStartTime,
-      Instant workHoursEndTime) {
+  public Collection<ScheduledTask> schedule(Collection<CalendarEvent> events,
```
jamesnam-google: Consider declaring one param per line.

ex:
...schedule(
    Collection...
    Collection...

portfolio/src/main/java/com/google/sps/servlets/ScheduleServlet.java
```diff
@@ -40,13 +41,32 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+    String workHoursStartTimeString = jsonFromRequest.getString("startTime");
+    String workHoursEndTimeString = jsonFromRequest.getString("endTime");
     Instant workHoursStartTime = Instant.parse(workHoursStartTimeString);
     Instant workHoursEndTime = Instant.parse(workHoursEndTimeString);
-
+    String algorithmTypeString = jsonFromRequest.getString("algorithmType");
     Collection<CalendarEvent> events = collectEventsFromJsonArray(eventsArray);
     Collection<Task> tasks = collectTasksFromJsonArray(tasksArray);
+
+    // These are null for initialization purposes. They will either be set or
```
jamesnam-google: I don't think the program should throw an error in case of error. The servlet can return some error response, like 400 (Bad Request). And this servlet should complete gracefully.

portfolio/src/main/java/com/google/sps/servlets/ScheduleServlet.java
```diff
@@ -40,13 +41,32 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+    // the program will thrown an error.
+    SchedulingAlgorithmType schedulingAlgorithmType = null;
+    BaseTaskScheduler algorithm = null;
+
+    // This will take the string representation of the algorithm that is passed
+    // in and match it to something in the enum under BaseTaskScheduler.
+    try {
```
jamesnam-google: Consider using a helper method that returns an optional containing SchedulingAlgorithmType.

ex:

Optional<SchedulingAlgorithmType> schedulingAlgorithmTypeOptional =
    getSchedulingAlgorithmTypeOptional(...);
if (!schedulingAlgorithmTypeOptional.isPresent()) {
  // Return a 400 response to the client and return the method.
}

Optional<BaseTaskScheduler> algorithmOptional = getAlgorithmOptional(...);
if (!algorithmOptional.isPresent()) {
  // Same thing here
}

...

portfolio/src/test/java/com/google/sps/ShortestTaskFirstTest.java
```diff
@@ -0,0 +1,187 @@
...
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public final class ShortestTaskFirstTest {
+  private static Instant BEGINNING_OF_DAY = Instant.parse("2020-06-25T00:00:00Z");
+  private static Instant END_OF_DAY = Instant.parse("2020-06-25T23:59:59Z");
+  private static Instant THREE_DAYS_LATER = Instant.parse("2020-06-28T00:00:00Z");
+
+  private static Duration TIME_15_MINUTES = Duration.ofSeconds(900);
```
jamesnam-google: Consider replacing 900 with 15 * 60 for readability. Here and below.

portfolio/src/test/java/com/google/sps/ShortestTaskFirstTest.java
jamesnam-google: Yes please use javadoc style in test methods.

portfolio/src/test/java/com/google/sps/ShortestTaskFirstTest.java
```diff
@@ -0,0 +1,187 @@
...
+    Collection<Task> tasks = Arrays.asList(task1);
+    ScheduledTask scheduledTask1 = new ScheduledTask(task1, TIME_0900);
+
+    Collection<ScheduledTask> actual =
+        algorithm.schedule(events, tasks, TIME_0900, THREE_DAYS_LATER);
+    Collection<ScheduledTask> expected = Arrays.asList(scheduledTask1);
+    Assert.assertEquals(expected, actual);
+  }
+
+  // Makes sure no errors are thrown if we have more tasks than we can possibly
```
jamesnam-google: You can consider including this info in the method name.

ex:
tooManyTasksToSchedule_shouldNotThrowError() {

portfolio/src/main/java/com/google/sps/data/ShortestTaskFirst.java
```diff
@@ -21,13 +21,9 @@
 import java.util.Comparator;
 import java.util.List;
 
-// TODO(tomasalvarez): Refactor this class into an abstract class where each of
-// the scheduling algorithms can be their own class that extends this one.
-// TODO: Add test coverage for scheduling algorithms
-
-public class FindSchedule {
+public class ShortestTaskFirst extends BaseTaskScheduler {
```
jamesnam-google: Consider using suffix to indicate that this is a task scheduler.

ex:
ShortestTaskFirstScheduler

portfolio/src/test/java/com/google/sps/ShortestTaskFirstTest.java
```diff
@@ -0,0 +1,187 @@
...
+  private static Instant TIME_1600 = TIME_1500.plus(TIME_60_MINUTES);
+  private static Instant TIME_1700 = TIME_1600.plus(TIME_60_MINUTES);
+  private static Instant TIME_1800 = TIME_1700.plus(TIME_60_MINUTES);
+
+  private final TaskPriority LOW_PRIORITY = new TaskPriority(1);
+
+  // Makes sure we return an empty list in the case where no tasks are passed.
+  @Test
+  public void noTasksScheduled() {
+    ShortestTaskFirst algorithm = new ShortestTaskFirst();
```
jamesnam-google: Consider renaming the variable to shortestTaskFirstScheduler. I think algorithm is a vague name.

portfolio/src/test/java/com/google/sps/ShortestTaskFirstTest.java
jamesnam-google: Also, I think the Duration variables can be named DURATION_* instead of TIME_*. That way we can easily distinguish duration variables from instant variables.

jamesnam-google (changes requested):

portfolio/src/main/java/com/google/sps/servlets/ScheduleServlet.java
rgcase: I think we can skip the `schedulingAlgorithmType` variable too since it's only used here, and like James said just write a `getAlgorithmOptional(...)` method that takes in the String.

portfolio/src/main/java/com/google/sps/data/BaseTaskScheduler.java
talvarez0507: Removed the TODO. That was an old TODO for myself but I included the tests already in this PR.

portfolio/src/main/java/com/google/sps/data/BaseTaskScheduler.java
talvarez0507: I don't think I can do that because I use this class in the schedule servlet which is in a different package.

talvarez0507: Done.

portfolio/src/main/java/com/google/sps/data/BaseTaskScheduler.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/data/BaseTaskScheduler.java
talvarez0507: Similar to my other comment, the servlet uses this method so I could not make it private.

portfolio/src/main/java/com/google/sps/data/ScheduledTask.java
talvarez0507: This code came almost directly from a week 5 walk-through. This method is mainly to simplify the logic for the other .equals(). If I remove this then instead of `&& equals(this, (ScheduledTask) other)` in the first .equals(), I would need something like:
```
return other instanceof ScheduledTask
        && this.task.equals(((ScheduledTask) b).task)
        && this.startTime.equals(((ScheduledTask) b).startTime);
```
Do you recommend I leave this method in or remove it?

portfolio/src/main/java/com/google/sps/data/ShortestTaskFirst.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/data/ShortestTaskFirst.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/servlets/ScheduleServlet.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/servlets/ScheduleServlet.java
talvarez0507: Done.

portfolio/src/test/java/com/google/sps/ShortestTaskFirstTest.java
talvarez0507: Done.

portfolio/src/test/java/com/google/sps/ShortestTaskFirstTest.java
talvarez0507: Done.

portfolio/src/test/java/com/google/sps/ShortestTaskFirstTest.java
talvarez0507: Done.

portfolio/src/test/java/com/google/sps/ShortestTaskFirstTest.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/data/BaseTaskScheduler.java
```diff
@@ -0,0 +1,30 @@
...
+// limitations under the License.
+
+package com.google.sps.data;
+
+import java.time.Instant;
+import java.util.Collection;
+
+public abstract class BaseTaskScheduler {
+  /** This is the scheduling method that all algorithm classes should implement. */
+  public abstract Collection<ScheduledTask> schedule(Collection<CalendarEvent> events,
```
jamesnam-google: Consider declaring one param per line.

portfolio/src/main/java/com/google/sps/servlets/ScheduleServlet.java
```diff
@@ -91,4 +117,35 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
   }
+
+  private static Optional<SchedulingAlgorithmType> getSchedulingAlgorithmTypeOptional(
+      String algorithmTypeString) {
+    // Here we should add a case for each new algorithm that is implemented.
+    switch (algorithmTypeString) {
+      case "SHORTEST_TASK_FIRST":
+        return Optional.of(SchedulingAlgorithmType.SHORTEST_TASK_FIRST);
+    }
+    return null;
```
jamesnam-google: I think you want to return Optional.empty() instead of null.

portfolio/src/main/java/com/google/sps/servlets/ScheduleServlet.java
```diff
@@ -91,4 +117,35 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+    // This will always be present because in the doPost we return the method
+    // before the code gets to call this method if this Optional is not
+    // present.
+    SchedulingAlgorithmType schedulingAlgorithmType = schedulingAlgorithmTypeOptional.get();
+    // Here we should add a case for each new algorithm that is implemented.
+    switch (schedulingAlgorithmType) {
+      case SHORTEST_TASK_FIRST:
+        return Optional.of(new ShortestTaskFirstScheduler());
+    }
+    return null;
```
jamesnam-google: Return Optional.empty()

portfolio/src/main/java/com/google/sps/servlets/ScheduleServlet.java
```diff
@@ -91,4 +117,35 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+      String algorithmTypeString) {
+    // Here we should add a case for each new algorithm that is implemented.
+    switch (algorithmTypeString) {
+      case "SHORTEST_TASK_FIRST":
+        return Optional.of(SchedulingAlgorithmType.SHORTEST_TASK_FIRST);
+    }
+    return null;
+  }
+
+  private static Optional<BaseTaskScheduler> getAlgorithmOptional(
```
jamesnam-google: Consider renaming to getTaskSchedulerOptional

portfolio/src/main/java/com/google/sps/data/BaseTaskScheduler.java
jamesnam-google: I think you don't want to reference BaseTaskScheduler outside this package. What you can do is create an interface that is public, with the two methods. Then you can have the task scheduler classes implement that interface, and reference the interface elsewhere rather than this abstract class.

portfolio/src/main/java/com/google/sps/servlets/ScheduleServlet.java
```diff
@@ -40,27 +43,50 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+    String algorithmTypeString = jsonFromRequest.getString("algorithmType");
     Collection<CalendarEvent> events = collectEventsFromJsonArray(eventsArray);
     Collection<Task> tasks = collectTasksFromJsonArray(tasksArray);
+
+    Optional<SchedulingAlgorithmType> schedulingAlgorithmTypeOptional =
+        getSchedulingAlgorithmTypeOptional(algorithmTypeString);
+    if (!schedulingAlgorithmTypeOptional.isPresent()) {
+      response.sendError(
+          HttpServletResponse.SC_BAD_REQUEST,
+          "The request by the client was syntactically incorrect.");
```
jamesnam-google: Consider clarifying that the algorithm type could not be determined.

portfolio/src/main/java/com/google/sps/servlets/ScheduleServlet.java
```diff
@@ -40,27 +43,50 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+      // Here I am returning the empty schedule instead of null to not mess up
+      // any front end code expecting some array.
+      returnEmptyArrayResponse(response);
+      return;
+    }
+
+    Optional<BaseTaskScheduler> algorithmOptional =
```
jamesnam-google: Consider renaming to taskSchedulerOptional. And as I suggested above, this variable can be of type Optional<TaskScheduler> where TaskScheduler is a new interface.

portfolio/src/main/java/com/google/sps/servlets/ScheduleServlet.java
```diff
@@ -40,27 +43,50 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+        getAlgorithmOptional(schedulingAlgorithmTypeOptional);
+    if (!algorithmOptional.isPresent()) {
+      response.sendError(
+          HttpServletResponse.SC_BAD_REQUEST,
+          "The request by the client was syntactically incorrect.");
```
jamesnam-google: Consider clarifying that the task scheduler could not be determined.

jamesnam-google (changes requested):

portfolio/src/main/java/com/google/sps/data/ScheduledTask.java
jamesnam-google: You can leave it here if you want.

portfolio/src/main/java/com/google/sps/data/BaseTaskScheduler.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/servlets/ScheduleServlet.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/servlets/ScheduleServlet.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/servlets/ScheduleServlet.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/servlets/ScheduleServlet.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/servlets/ScheduleServlet.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/servlets/ScheduleServlet.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/data/BaseTaskScheduler.java
talvarez0507: In hindsight, some of the stuff I'm doing in this servlet makes it impossible to have schedulers be package-private. As a result I created a class called ServletUtils  to 1. clean up the servlet a bit 2. allow the schedulers to be package-private. All I did was copy over the existing helper methods and put them into that class.

portfolio/src/main/java/com/google/sps/data/ServletUtils.java
```diff
@@ -0,0 +1,92 @@
...
+        int priorityInt = taskJsonObject.getInt("taskPriority");
+        TaskPriority priority = new TaskPriority(priorityInt);
+        Task newTask = new Task(name, description, duration, priority);
+        tasks.add(newTask);
+      }
+    }
+    return tasks;
+  }
+
+  public static Optional<SchedulingAlgorithmType> getSchedulingAlgorithmTypeOptional(
```
jamesnam-google: I think this method and the one below can be in a separate file (similar to this class), like SchedulingAlgorithmReference.java. I usually try to avoid naming classes Util or Utils, because they are a bit ambiguous. But I'm fine with keeping this class named as ServletUtils (or maybe ServletHelper).

jamesnam-google (approved):
Approving the change with a comment.

portfolio/src/main/java/com/google/sps/data/ServletUtils.java
talvarez0507: Done.

rgcase (approved):

PR #10, opened 2020-07-08
Client-side request and UI elements

talvarez0507:
This PR includes the HTML elements for displaying the scheduled results. It also includes the JS code that scrapes the information off of the UI and fetches information from the server.

### DIFF
--- a/portfolio/src/main/webapp/request.js
+++ b/portfolio/src/main/webapp/request.js
@@ -33,14 +33,17 @@ var scheduledTasks = [];
  * Gets all of the scheduling information from the UI and returns a
  * ScheduleRequest with all of the data.
  */
-function createRequestFromUiInformation() {
+function createScheduleRequestFromDom() {
   const startTime = document.getElementById('working-hour-start').value;
   const endTime = document.getElementById('working-hour-end').value;
   const events = collectAllEvents();
   const tasks = collectAllTasks();
   const algorithmType = document.getElementById('algorithm-type').value;
   const scheduleRequest = new ScheduleRequest(
-      events, tasks, getTimeObject(startTime), getTimeObject(endTime),
+      events,
+      tasks,
+      getTimeObject(startTime),
+      getTimeObject(endTime),
       algorithmType);
   return scheduleRequest;
 }
@@ -48,22 +51,22 @@ function createRequestFromUiInformation() {
 /**
  * Gets called when the user hits 'Start Scheduling'.
  */
-function startScheduling() {
-  const scheduleRequest = createRequestFromUiInformation();
+function onClickStartScheduling() {
+  const scheduleRequest = createScheduleRequestFromDom();
   // Create the request to send to the server using the data we collected from
   // the web form.
   fetchScheduledTasksFromServlet(scheduleRequest).then((scheduledTaskArray) => {
-    updateResultsOnPage(scheduledTaskArray);
+    handleScheduledTaskArray(scheduledTaskArray);
   });
 }
 
 /**
  * Updates the UI to show the results of a query.
  */
-function updateResultsOnPage(scheduledTaskArray) {
+function handleScheduledTaskArray(scheduledTaskArray) {
   const resultElement = document.getElementById('scheduled-task-list');
   resultElement.innerHTML = '';
-  scheduledTaskArray.forEach(addScheduledTaskToPage, resultElement);
+  scheduledTaskArray.forEach(addScheduledTaskToDom, resultElement);
 }
 
 /**
@@ -71,15 +74,11 @@ function updateResultsOnPage(scheduledTaskArray) {
  * scheduled tasks.
  */
 function fetchScheduledTasksFromServlet() {
-  const scheduleRequest = createRequestFromUiInformation();
+  const scheduleRequest = createScheduleRequestFromDom();
   const json = JSON.stringify(scheduleRequest);
   return fetch('/schedule', {method: 'POST', body: json})
       .then((response) => {
-        // This is still a Promise.
-        return response.json();
-      })
-      // Turns result from a Promise into its Array value.
-      .then((array) => {
+        var array = response.json();
         scheduledTasks = array;
         return scheduledTasks;
       });
@@ -96,7 +95,7 @@ function collectAllScheduledTasks() {
  * This takes in a single JSON object of a ScheduledTask (Java Class) and
  * displays that information on the result element using a card format.
  */
-function addScheduledTaskToPage(scheduledTaskWrapper, resultElement) {
+function addScheduledTaskToDom(scheduledTaskWrapper, resultElement) {
   const task = scheduledTaskWrapper.task;
   const taskName = task.name;
   const taskTimeSeconds = scheduledTaskWrapper.startTime.seconds;
--- a/portfolio/src/main/webapp/request.js
+++ b/portfolio/src/main/webapp/request.js
@@ -74,7 +74,6 @@ function collectAllScheduledTasks() {
   });
 }
 
-
 /**
  * This takes in a single JSON object of a ScheduledTask (Java Class) and
  * displays that information on the result element using a card format.
@@ -106,7 +105,6 @@ function addScheduledTaskToPage(scheduledTaskWrapper, resultElement) {
   timeText.classList.add('card-text');
   timeText.innerText = 'Scheduled for: ' + taskDate;
   cardBody.appendChild(timeText);
-
 
   const eventList = document.getElementById('scheduled-task-list');
   eventList.innterHTML = '';
--- a/portfolio/src/main/webapp/request.js
+++ b/portfolio/src/main/webapp/request.js
@@ -9,8 +9,8 @@ class ScheduleRequest {
    * All fields are required for this object.
    * @param events: events happening the day of the scheduling. This is an
    *     array of JavaScript CalendarEvent objects.
-   * @param tasks: the tasks a user would like to schedule on that day. This is an
-   *     array of JavaScript Task objects.
+   * @param tasks: the tasks a user would like to schedule on that day. This is
+   *     an array of JavaScript Task objects.
    * @param startTime: a JavaScript Date object representing the start time.
    * @param endTime: a JavaScript Date object representing the end time.
    * @param algorithmType: a string representation of the algorithm type that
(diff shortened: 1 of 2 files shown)


portfolio/src/main/webapp/request.js
```diff
@@ -25,4 +25,88 @@ class ScheduleRequest {
...
+}
+
+/**
+ * Gets called when the user hits 'Start Scheduling'.
+ */
+function startScheduling() {
+  const scheduleRequest = createRequestFromUiInformation();
+  // Create the request to send to the server using the data we collected from
+  // the web form.
+  collectAllScheduledTasks(scheduleRequest).then((scheduledTaskArray) => {
```
hollyyuqizheng: Do you think if you could store the scheduledTaskArray into its variable once it is received from the server, and return this array somehow? The collectAllScheduledTasks method does this, but I don't think it's necessary to send a request to the server again after the scheduling is done and the scheduled tasks are stored somewhere. I think having a getter method that returns the scheduledTaskArray will be helpful for updating and exporting to Calendar and Tasks later.

portfolio/src/main/webapp/request.js
```diff
@@ -25,4 +25,88 @@ class ScheduleRequest {
...
+function addScheduledTaskToPage(scheduledTaskWrapper, resultElement) {
+  const task = scheduledTaskWrapper.task;
+  const taskName = task.name;
+  const taskTimeSeconds = scheduledTaskWrapper.startTime.seconds;
+  const taskDate = new Date();
+  // This is x1000 because the functions takes milliseconds
+  taskDate.setTime(taskTimeSeconds * 1000);
+  const taskPriority = task.priority.priority;
+  resultElement.innerText += taskName;
+  resultElement.innerText += taskName;
```
hollyyuqizheng: Is adding the name twice necessary?

hollyyuqizheng (changes requested):
Suggestions on more methods to add that might be helpful for other parts of the UI.

portfolio/src/main/webapp/request.js
talvarez0507: I broke the up functionality and added a new local variable scheduledTasks. One function will now fetch from the servlet and set the local variable while another function simply returns this local variable.

portfolio/src/main/webapp/request.js
talvarez0507: Removed.

portfolio/src/main/webapp/request.js
```diff
@@ -25,4 +25,109 @@ class ScheduleRequest {
...
-// TODO(tomasalvarez): write function to fetch results and display results
+// Once a call is made to the server for scheduling this array will be a local
+// copy of the results.
+var scheduledTasks = []
+
+    /**
+     * Gets all of the scheduling information from the UI and returns a
+     * ScheduleRequest with all of the data.
+     */
+    function
```
hollyyuqizheng: nit: indentation

portfolio/src/main/webapp/request.js
```diff
@@ -25,4 +25,109 @@ class ScheduleRequest {
...
+
+/**
+ * Handles sending the ScheduleRequest object to the servlet and returns the
+ * scheduled tasks.
+ */
+function
+fetchScheduledTasksFromServlet() {
+  const scheduleRequest = createRequestFromUiInformation();
+  const json = JSON.stringify(scheduleRequest);
+  return fetch('/schedule', {method: 'POST', body: json})
```
hollyyuqizheng, line 79: I wonder if you need this return on this line? Would the final `return response.json()` return the Json object there?

hollyyuqizheng (approved):

portfolio/src/main/webapp/request.js
```diff
@@ -25,4 +25,109 @@ class ScheduleRequest {
...
+          events, tasks, getTimeObject(startTime), getTimeObject(endTime),
+          algorithmType);
+      return scheduleRequest;
+    }
+
+/**
+ * Gets called when the user hits 'Start Scheduling'.
+ */
+function
+startScheduling() {
```
raulcruise: Unnecessary line break

portfolio/src/main/webapp/request.js
```diff
@@ -25,4 +25,109 @@ class ScheduleRequest {
...
+  resultElement.innerHTML = '';
+  scheduledTaskArray.forEach(addScheduledTaskToPage, resultElement);
+}
+
+/**
+ * Handles sending the ScheduleRequest object to the servlet and returns the
+ * scheduled tasks.
+ */
+function
+fetchScheduledTasksFromServlet() {
```
raulcruise: Again do we need a line break here? It seems that the comment above is longer than this line would be.

portfolio/src/main/webapp/request.js
```diff
@@ -25,4 +25,109 @@ class ScheduleRequest {
...
+        scheduledTasks = array;
+        return scheduledTasks;
+      });
+}
+
+/**
+ * Returns the local copy of the latest scheduled tasks.
+ */
+function
+collectAllScheduledTasks() {
```
raulcruise: Here too

portfolio/src/main/webapp/request.js
talvarez0507: I kind of forgot if this needs to be a return or not but the walkthroughs put a return here so I did too.

portfolio/src/main/webapp/request.js

portfolio/src/main/webapp/request.js

portfolio/src/main/webapp/request.js

portfolio/src/main/webapp/request.js

raulcruise (approved):

portfolio/src/main/webapp/request.js
```diff
@@ -25,4 +25,105 @@ class ScheduleRequest {
...
+ * ScheduleRequest with all of the data.
+ */
+function createRequestFromUiInformation() {
+  const startTime = document.getElementById('working-hour-start').value;
+  const endTime = document.getElementById('working-hour-end').value;
+  const events = collectAllEvents();
+  const tasks = collectAllTasks();
+  const algorithmType = document.getElementById('algorithm-type').value;
+  const scheduleRequest = new ScheduleRequest(
+      events, tasks, getTimeObject(startTime), getTimeObject(endTime),
```
jamesnam-google: Consider declaring one param per line

portfolio/src/main/webapp/request.js
```diff
@@ -25,4 +25,105 @@ class ScheduleRequest {
...
-// TODO(tomasalvarez): write function to fetch results and display results
+// Once a call is made to the server for scheduling this array will be a local
+// copy of the results.
+var scheduledTasks = [];
+
+/**
+ * Gets all of the scheduling information from the UI and returns a
+ * ScheduleRequest with all of the data.
+ */
+function createRequestFromUiInformation() {
```
jamesnam-google: Consider renaming to

createScheduleRequestFromDom

portfolio/src/main/webapp/request.js
```diff
@@ -25,4 +25,105 @@ class ScheduleRequest {
...
+  const scheduleRequest = new ScheduleRequest(
+      events, tasks, getTimeObject(startTime), getTimeObject(endTime),
+      algorithmType);
+  return scheduleRequest;
+}
+
+/**
+ * Gets called when the user hits 'Start Scheduling'.
+ */
+function startScheduling() {
```
jamesnam-google: Consider renaming to onClickStartScheduling

portfolio/src/main/webapp/request.js
```diff
@@ -25,4 +25,105 @@ class ScheduleRequest {
...
+  // the web form.
+  fetchScheduledTasksFromServlet(scheduleRequest).then((scheduledTaskArray) => {
+    updateResultsOnPage(scheduledTaskArray);
+  });
+}
+
+/**
+ * Updates the UI to show the results of a query.
+ */
+function updateResultsOnPage(scheduledTaskArray) {
```
jamesnam-google: Consider renaming to handleScheduledTaskArray

portfolio/src/main/webapp/request.js
```diff
@@ -25,4 +25,105 @@ class ScheduleRequest {
...
+ * scheduled tasks.
+ */
+function fetchScheduledTasksFromServlet() {
+  const scheduleRequest = createRequestFromUiInformation();
+  const json = JSON.stringify(scheduleRequest);
+  return fetch('/schedule', {method: 'POST', body: json})
+      .then((response) => {
+        // This is still a Promise.
+        return response.json();
+      })
```
jamesnam-google: I don't think it's necessary to use a future to chain. Can we do something like,

.then((response) => {
  var array = response.json();
  scheduledTasks = array;
  return scheduledTasks;
}

portfolio/src/main/webapp/request.js
```diff
@@ -25,4 +25,105 @@ class ScheduleRequest {
...
+ */
+function collectAllScheduledTasks() {
+  return scheduledTasks;
+}
+
+/**
+ * This takes in a single JSON object of a ScheduledTask (Java Class) and
+ * displays that information on the result element using a card format.
+ */
+function addScheduledTaskToPage(scheduledTaskWrapper, resultElement) {
```
jamesnam-google: Consider renaming to

addScheduledTaskToDom

portfolio/src/main/webapp/request.js
```diff
@@ -25,4 +25,105 @@ class ScheduleRequest {
   }
 }
 
-// TODO(tomasalvarez): write function to fetch results and display results
+// Once a call is made to the server for scheduling this array will be a local
+// copy of the results.
+var scheduledTasks = [];
```
jamesnam-google, line 30: Is this local variable necessary? I think you render the scheduled tasks on the DOM, right? So you may not need this local variable. If we have a variable like this then it can be burdensome to keep it synced with the DOM.

jamesnam-google (approved):
Approving the change with comments.

portfolio/src/main/webapp/request.js
rgcase: Ya you still need to `return` here, because the result of this function is that you're returning a promise to the caller. You can think of a promise just like any other object that needs to be returned. It's exactly the same as if say you wanted to transform and return an array of something like:
```
function foo() {
  var array = [1, 2, 3];
  return array.map((i) => { return i + 1; });
}
```
returns `[2,3,4]`.

rgcase (approved):

portfolio/src/main/webapp/request.js
talvarez0507: Currently this gets updated when you hit "Start Scheduling". I'd like to make the scheduled task card include all of the information (not just name and time) and then I will remove this local variable and get the information from the dom. That will probably happen in a PR soon.

portfolio/src/main/webapp/request.js

portfolio/src/main/webapp/request.js

portfolio/src/main/webapp/request.js

portfolio/src/main/webapp/request.js

portfolio/src/main/webapp/request.js

portfolio/src/main/webapp/request.js

PR #11, opened 2020-07-09
Initial upload of API key servlet

hollyyuqizheng:
To use this servlet, I first had to export a variable called "API_KEY" in my .bashrc file. Then I added a function in calendarApi.js to make a GET request to this servlet. The servlet will fetch the API key from the environment variable and send it back to the client side.


jamesnam-google (approved):

portfolio/src/main/java/com/google/sps/servlets/ApiKeyServlet.java
```diff
@@ -0,0 +1,20 @@
...
+import javax.servlet.http.HttpServletResponse;
+
+@WebServlet("/apiKeyServlet")
+public class ApiKeyServlet extends HttpServlet {
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    Map<String, String> env = System.getenv();
+    String apiKey = env.get("API_KEY");
+    response.setContentType("text/html");
```
rgcase: Is it really html? Maybe we should have some json object containing the key to be more explicit?

rgcase (changes requested):

raulcruise (approved):

talvarez0507 (approved):

PR #12, opened 2020-07-09
Uploading API Key servlet

hollyyuqizheng:
To use this servlet, I first had to export a variable called "API_KEY" in my .bashrc file. Then I added a function in calendarApi.js to make a GET request to this servlet. The servlet will fetch the API key from the environment variable and send it back to the client side.


rgcase (approved):

talvarez0507 (approved):

portfolio/src/main/java/com/google/sps/servlets/ApiKeyServlet.java
```diff
@@ -0,0 +1,30 @@
...
+import com.google.gson.Gson;
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+@WebServlet("/apiKeyServlet")
```
jamesnam-google: I think this servlet can be used in the future to deliver other key-value entries to the client. So consider making it more general, like /appConfigServlet or something. Then we can potentially use it for other config fields, in addition to the API key.

jamesnam-google (approved):
Approving the change with comment.

raulcruise (approved):

PR #13, opened 2020-07-10
Add Mocha tests to task.js

raulcruise:
Use Mocha to test standalone functions within task.js.


talvarez0507 (approved):

hollyyuqizheng (approved):

portfolio/src/main/webapp/task.js
```diff
@@ -174,3 +174,7 @@ function validateTaskDuration(duration) {
     return {isValid: true, errorMessage: null};
   }
 }
+
+module.exports.getDurationMinutes = getDurationMinutes;
```
jamesnam-google, line 179: Consider creating a Javascript class for duration. I think Date works for instant, but I don't see a counter-part for Duration Java class.

To begin with, the Duration class can just be instantiated with milliseconds via constructor. And you can start using the Duration class in the code-base.

But if you want to handle this later on in the internship that's fine too. Consider adding TODO comment.

jamesnam-google (approved):
Approving the change with comment

rgcase (approved):

PR #17, opened 2020-07-11
Add Function to clear Tasks from UI

raulcruise:
This function would be very useful in the case that the user imports all of their Tasks from the Tasks API and decides he would like to undo it.

### DIFF
--- a/portfolio/src/main/webapp/index.html
+++ b/portfolio/src/main/webapp/index.html
@@ -129,7 +129,7 @@
           <div class="scrollable" id="new-task-list-wrapper">
             <div id="new-task-list"></div>
           </div>
-          <button type="button" class="btn btn-primary" onclick="clearTasks()">Clear Tasks</button>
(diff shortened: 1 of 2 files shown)


hollyyuqizheng (approved):

talvarez0507 (approved):

portfolio/src/main/webapp/index.html
```diff
@@ -129,6 +129,7 @@ <h2>Tasks added:</h2>
           <div class="scrollable" id="new-task-list-wrapper">
             <div id="new-task-list"></div>
           </div>
+          <button type="button" class="btn btn-primary" onclick="clearTasks()">Clear Tasks</button>
```
jamesnam-google: Consider using non-primary class names, like btn-warning or btn-danger.

jamesnam-google (approved):

rgcase (approved):

portfolio/src/main/webapp/index.html
raulcruise: Done.

PR #18, opened 2020-07-14
Fix error in request.js

talvarez0507:
I incorrectly handled some .then() statements and only now realized that the code I merged has a bug. This PR fixes that bug.

### DIFF
--- a/portfolio/src/main/webapp/request.js
+++ b/portfolio/src/main/webapp/request.js
@@ -55,7 +55,7 @@ function onClickStartScheduling() {
   const scheduleRequest = createScheduleRequestFromDom();
   // Create the request to send to the server using the data we collected from
   // the web form.
-  fetchScheduledTasksFromServlet(scheduleRequest).then((scheduledTaskArray) => {
+  fetchScheduledTasksFromServlet().then((scheduledTaskArray) => {
     scheduledTasks = scheduledTaskArray;
     handleScheduledTaskArray(scheduledTaskArray);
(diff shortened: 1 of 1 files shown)


hollyyuqizheng (approved):

raulcruise (approved):

portfolio/src/main/webapp/request.js
```diff
@@ -55,7 +55,8 @@ function onClickStartScheduling() {
   const scheduleRequest = createScheduleRequestFromDom();
   // Create the request to send to the server using the data we collected from
   // the web form.
-  fetchScheduledTasksFromServlet(scheduleRequest).then((scheduledTaskArray) => {
+  fetchScheduledTasksFromServlet().then((scheduledTaskArray) => {
+    scheduledTasks = scheduledTaskArray;
```
jamesnam-google: I think I mentioned this before, but do we need this scheduledTasks variable? Can it diverge from the DOM elements? It might be safer/simpler to treat DOM elements as the source of truth.

jamesnam-google (approved):

rgcase:
> I incorrectly handled some .then() statements and only now realized that the code I merged has a bug. This PR fixes that bug.

Hey Tomas. It's not really clear to me why this was a bug and how your PR fixes it. Can you leave a little more information about the bug, either as a comment here or in a maybe in a buganizer bug with the details?

rgcase (approved):

PR #19, opened 2020-07-15
TimeRangeGroup interface

hollyyuqizheng:
The TimeRangeGroup interface models a collection of time ranges. The interface contains methods for adding, merging, and deleting time ranges. This interface is potentially useful to model a group of free time ranges and the modifications that are done to a group of free time ranges during the scheduling algorithm.


portfolio/src/test/java/com/google/sps/TimeRangeGroupTest.java
```diff
@@ -0,0 +1,224 @@
...
+
+    List<TimeRange> expectedTimeRangesAfterDelete = Arrays.asList(timeRangeTwo, timeRangeThree);
+    List<TimeRange> actualTimeRangesAfterDelete =
+        timeRangeGroupOriginal.deletePartOfTimeRange(timeRangeOne, timeRangeToDelete);
+
+    Assert.assertEquals(expectedTimeRangesAfterDelete, actualTimeRangesAfterDelete);
+  }
+
+  /**
+   * Tests deleting from a time range but withou the wrong boundary. The target time range to delete
```
raulcruise: nit: without*

raulcruise (approved):

portfolio/src/test/java/com/google/sps/TimeRangeGroupTest.java
hollyyuqizheng: Updated. It should be "with" actually.

portfolio/src/main/java/com/google/sps/data/TimeRangeGroup.java
```diff
@@ -0,0 +1,97 @@
...
+    return allTimeRanges;
+  }
+
+  /**
+   * Deletes only part of a time range from the middle.
+   *
+   * @return the sorted all time ranges with the new ranges included.
+   */
+  public List<TimeRange> deletePartOfTimeRange(
+      TimeRange originalTimeRange, TimeRange timeRangeToDelete) {
```
talvarez0507: These variable names are seem inconsistent. I think either one of the pairs `originalTimeRange, toDeleteTimeRange` or `timeRangeOriginal, timeRangeToDelete` would be more consistent.

portfolio/src/main/java/com/google/sps/data/TimeRangeGroup.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+    return allTimeRanges;
+  }
+
+  /**
+   * Deletes only part of a time range from the middle.
+   *
+   * @return the sorted all time ranges with the new ranges included.
+   */
+  public List<TimeRange> deletePartOfTimeRange(
```
talvarez0507: (Not a change request just ideas that we can discuss with James and Ryan) In my 1 on 1 with James we discussed a very similar problem regarding DS/Algo complexity. It was mentioned that something like a Linked List here would be good so instead of having to sort before returning we could insert the TimeRange's quickly. Another idea could be the very same method but the only argument is timeRangeToDelete. Instead of looping through and finding the right TimeRange that coincides with toDelete there can be some pre-processing data structure so that the search is quicker.

talvarez0507 (approved):

portfolio/src/main/java/com/google/sps/data/TimeRangeGroup.java

portfolio/src/main/java/com/google/sps/data/TimeRangeGroup.java
hollyyuqizheng: I see, will try!

portfolio/src/main/java/com/google/sps/data/TimeRangeGroup.java
```diff
@@ -0,0 +1,97 @@
+package com.google.sps.data;
+
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+/** Models a collection of Time Ranges with methods for add and delete. */
+public class TimeRangeGroup {
```
jamesnam-google: I have a few suggestions.

First, I would suggest that you use an interface, which can have multiple implementations. The interface can be named TimeRangeGroup, and maybe the initial implementation can be named like ArrayListTimeRangeGroup, given that it's backed by an array-list internally. But in the future there can be other implementations of the interface, like LinkedListTimeRangeGroup or BinaryTreeTimeRangeGroup.

Here are the methods that I see currently, for reference:
public void addTimeRange(TimeRange newTimeRange);
public boolean equals(Object other);
public List<TimeRange> getAllTimeRanges();
public List<TimeRange> deleteEntireTimeRange(TimeRange timeRange);
public List<TimeRange> deletePartOfTimeRange(
    TimeRange timeRangeOriginal, TimeRange timeRangeToDelete);

1. Given that you are trying to use this data structure to track the free time ranges, I think you may want to implement some merging behavior when it comes to addTimeRange. For example, suppose that there was one time range internally, like 1pm-2pm today. And if the user adds another time range from 1:30pm-4pm, then it's equivalent to one time range from 1pm-4pm (free time).

2. I think it can be worthwhile to add another method like hasTimeRange. If for example internally the class was keeping track of 1pm-2pm and 3pm-4pm, and hasTimeRange is called with 3pm-3:30pm, then it would return true. If it was called with 1pm-4pm, then it would return false, because there is a gap (2pm-3pm).

3. I think deleteEntireTimeRange and deletePartOfTimeRange can be merged into deleteTimeRange, which takes some time range. And the method can throw an exception if hasTimeRange is false. For example, suppose internally the class was keeping track of 1pm-2pm and 3pm-4pm. If delete time range is called with 1pm-2pm, then just 3pm-4pm remains. If the same method is called with 1:30pm-1:45pm, then we are left with 1pm-1:30pm, 1:45pm-2pm, and 3pm-4pm. This delete method can be used when a task gets scheduled.

4. I think getAllTimeRanges should return the merged, disjoint time ranges, probably sorted by time range start times. I don't think it should just return the time ranges that were added. For example, suppose two time ranges were added: 1pm-2pm, and then 10am-5pm. When getAllTimeRanges is called, it should return 10am-5pm, rather than [1pm-2pm, 10am-5pm].

portfolio/src/main/java/com/google/sps/data/TimeRangeGroup.java
```diff
@@ -0,0 +1,97 @@
...
+
+  public List<TimeRange> allTimeRanges;
+
+  public TimeRangeGroup() {
```
jamesnam-google: Consider adding a constructor that takes a collection of time ranges. And the constructor can perform the merging behavior mentioned above.

jamesnam-google (changes requested):
I left a couple comments.

talvarez0507 (approved):

raulcruise (approved):

portfolio/src/main/java/com/google/sps/data/TimeRangeGroup.java
```diff
@@ -0,0 +1,28 @@
...
+
+import java.util.Collection;
+
+public interface TimeRangeGroup {
+
+  /**
+   * Merges the collection of time ranges that is passed into the constructor so that all time
+   * ranges are disjoint.
+   */
+  public Collection<TimeRange> mergeInputTimeRanges(Collection<TimeRange> timeRanges);
```
jamesnam-google: Consider renaming to addTimeRanges. I think the merging should be done internally, rather than explicitly specified by the callers.

portfolio/src/main/java/com/google/sps/data/TimeRangeGroup.java
```diff
@@ -0,0 +1,28 @@
...
+
+  /** Adds a new time range into the collection of time ranges. */
+  public void addTimeRange(TimeRange timeRange);
+
+  /** Returns the collection of all time ranges. */
+  public Collection<TimeRange> getAllTimeRanges();
+
+  /** Checks if a time range exists in the collection. */
```
jamesnam-google: Consider clarifying the behavior. For example, if the time range group has 1pm-2pm and 3pm-4pm, and this method is called with 3pm-3:30pm, does this method return true? (I think it should.)

portfolio/src/main/java/com/google/sps/data/TimeRangeGroup.java
```diff
@@ -0,0 +1,28 @@
...
+  public boolean hasTimeRange(TimeRange timeRange);
+
+  /**
+   * Deletes a given time range.
```
jamesnam-google: Similarly, please clarify the method behavior.

jamesnam-google (approved):

portfolio/src/main/java/com/google/sps/data/TimeRangeGroup.java

portfolio/src/main/java/com/google/sps/data/TimeRangeGroup.java
hollyyuqizheng: Updated. There will not be merging logic in this interface, since all time ranges passed in are free time ranges calculated in the CalendarEventsGroup class, and these free time ranges will be disjoint.

portfolio/src/main/java/com/google/sps/data/TimeRangeGroup.java

rgcase (approved):

PR #21, opened 2020-07-15
Clear create task input on task addition

raulcruise:
Fulfilling feature request b/161290136

### DIFF
--- a/portfolio/src/main/webapp/task.js
+++ b/portfolio/src/main/webapp/task.js
@@ -158,13 +158,14 @@ function clearTasks() {
 }
 
 /**
- * This method clears empties the three text input boxes
- * on the create new task UI.
+ * This method clears inputs from the create new task UI
+ * once a user has added a Task by setting the input values
+ * to empty strings.
  */
 function clearNewTaskInputs() {
-  $('#new-task-estimated-length').empty();
-  $('#new-task-name').empty();
-  $('#new-task-description').empty();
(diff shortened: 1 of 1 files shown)


portfolio/src/main/webapp/task.js
```diff
@@ -153,6 +157,17 @@ function clearTasks() {
   taskList.innerHTML = '';
 }
 
+/**
+ * This method clears inputs from the create new task UI
+ * once a user has added a Task by setting the input values
+ * to empty strings.
+ */
+function clearNewTaskInputs() {
```
hollyyuqizheng, line 253: I think you could also use $('#new-task-name').empty() as well.

hollyyuqizheng (approved):

portfolio/src/main/webapp/task.js
raulcruise: Done.

raulcruise: .empty() would not work here because it empties the inner HTML but what we want it to empty the value since what is written into the input is not stored as: ```<input> User input goes here</input>``` but as ```<input value="User input goes here">```.

hollyyuqizheng: I see, thanks for letting me know!

talvarez0507 (approved):

jamesnam-google (approved):

rgcase (approved):

PR #25, opened 2020-07-17
ScheduledTask data now DOM based

talvarez0507:
This PR makes it so all of the ScheduledTask data is now displayed on the DOM. Additionally the function that retrieves the scheduledTasks now uses the DOM since the information is present there. A local variable storing all of the scheduledTasks was removed because it is no longer necessary. The Sheets API data collection was also updated to reflect the changes in how the data is presented.

### DIFF
--- a/portfolio/src/main/webapp/request.js
+++ b/portfolio/src/main/webapp/request.js
@@ -79,21 +79,18 @@ function collectAllScheduledTasks() {
   const scheduledTaskList = document.getElementById('schedule-result-list');
 
   scheduledTaskList.childNodes.forEach((scheduledTaskCard) => {
-    // The cardBody of the scheduledTaskCard
     const cardBody = scheduledTaskCard.childNodes[0];
 
-    const scheduledTaskName = cardBody.childNodes[0].innerText;
+    const scheduledTaskName =
+        cardBody.childNodes[0].getAttribute('data-task-name');
     const scheduledTaskScheduledTime =
-        cardBody.childNodes[1].innerText.substring(
-            SCHEDULED_TIME_INITIAL_TEXT.length);
+        cardBody.childNodes[1].getAttribute('data-task-time');
     const scheduledTaskDurationMinutes =
-        parseInt(cardBody.childNodes[2].innerText.substring(
-            DURATION_INITIAL_TEXT.length));
-    const scheduledTaskDescription = cardBody.childNodes[3].innerText.substring(
-        DESCRIPTION_INITIAL_TEXT.length);
+        parseInt(cardBody.childNodes[2].getAttribute('data-task-duration'));
+    const scheduledTaskDescription =
+        cardBody.childNodes[3].getAttribute('data-task-description');
     const scheduledTaskPriority =
-        parseInt(cardBody.childNodes[4].innerText.substring(
-            PRIORITY_INITIAL_TEXT.length));
+        parseInt(cardBody.childNodes[4].getAttribute('data-task-priority'));
 
     var scheduledTask = {};
     const task = new Task(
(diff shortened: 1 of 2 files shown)


hollyyuqizheng (approved):

raulcruise (approved):

portfolio/src/main/webapp/request.js
```diff
@@ -64,9 +65,46 @@ function onClickStartScheduling() {
...
+            DURATION_INITIAL_TEXT.length));
+    const scheduledTaskDescription = cardBody.childNodes[3].innerText.substring(
+        DESCRIPTION_INITIAL_TEXT.length);
+    const scheduledTaskPriority =
+        parseInt(cardBody.childNodes[4].innerText.substring(
+            PRIORITY_INITIAL_TEXT.length));
+
+    var scheduledTask = {};
+    const task = new Task(
+        scheduledTaskName, scheduledTaskDescription,
```
jamesnam-google: Consider placing one param per line. For example, when there are many params (like 7), often we end up looking for the 5th param. It's easier to find that when there is one param per line.

portfolio/src/main/webapp/request.js
```diff
@@ -64,9 +65,46 @@ function onClickStartScheduling() {
...
+function collectAllScheduledTasks() {
+  const allScheduledTasks = new Array();
+  // Get scheduled task list element to run through cards collecting data.
+  const scheduledTaskList = document.getElementById('schedule-result-list');
+
+  scheduledTaskList.childNodes.forEach((scheduledTaskCard) => {
+    // The cardBody of the scheduledTaskCard
+    const cardBody = scheduledTaskCard.childNodes[0];
+
+    const scheduledTaskName = cardBody.childNodes[0].innerText;
```
jamesnam-google: I would suggest that you store the data in some HTML element attribute, rather than inner text.

For example, the scheduled task name HTML element can have a data attribute, like,

data-task-name="..."

Then you don't have to do this substring operation. See the pages below for more info.

https://www.w3schools.com/tags/att_data-.asp
https://api.jquery.com/data/

jamesnam-google (changes requested):

portfolio/src/main/webapp/request.js

portfolio/src/main/webapp/request.js
talvarez0507: Done

portfolio/src/main/webapp/request.js
```diff
@@ -52,10 +54,9 @@ function createScheduleRequestFromDom() {
  * Gets called when the user hits 'Start Scheduling'.
  */
 function onClickStartScheduling() {
-  const scheduleRequest = createScheduleRequestFromDom();
   // Create the request to send to the server using the data we collected from
   // the web form.
-  fetchScheduledTasksFromServlet(scheduleRequest).then((scheduledTaskArray) => {
+  fetchScheduledTasksFromServlet().then((scheduledTaskArray) => {
```
rgcase: nit: I know it was already like this before, and it's up to you to decide what you prefer but I think you might be able to just pass the `handleScheduledTaskArray` function directly like this:
```
fetchScheduledTasksFromServlet().then(handleScheduledTaskArray);
```

portfolio/src/main/webapp/request.js
```diff
@@ -64,9 +65,46 @@ function onClickStartScheduling() {
...
+    const scheduledTaskPriority =
+        parseInt(cardBody.childNodes[4].getAttribute('data-task-priority'));
+
+    var scheduledTask = {};
+    const task = new Task(
+        scheduledTaskName,
+        scheduledTaskDescription,
+        scheduledTaskDurationMinutes,
+        scheduledTaskPriority
+        );
```
jamesnam-google: nit: This end paren can go to the end of the previous line.

ex:
scheduledTaskPriority);

rgcase: nit: no newline before `);`.

portfolio/src/main/webapp/request.js
```diff
@@ -115,14 +145,34 @@ function addScheduledTaskToDom(scheduledTaskWrapper, resultElement) {
...
   cardBody.appendChild(timeText);
 
-  const eventList = document.getElementById('scheduled-task-list');
-  eventList.innterHTML = '';
-  eventList.appendChild(newResultCard);
+
+  const durationText = document.createElement('p');
+  durationText.classList.add('card-text');
+  durationText.innerText = DURATION_INITIAL_TEXT + taskDurationMinutes;
+  durationText.setAttribute('data-task-duration', taskDurationMinutes);
```
jamesnam-google: Consider renaming the attribute name to data-task-duration-minutes so the time unit is clear.

rgcase (approved):

portfolio/src/main/webapp/sheetsApi.js
```diff
@@ -147,16 +147,13 @@ function singleScheduledTaskToSheetsArray(scheduledTask) {
...
   const taskName = task.name;
   // Changes seconds into minutes.
-  const taskDurationMinutes = task.duration.seconds / 60;
-  const taskDescription = task.description.value;
-  const taskPriority = task.priority.priority;
-  const taskTimeSeconds = scheduledTask.startTime.seconds;
-  const taskDate = new Date();
-  // This is x1000 because the functions takes milliseconds
-  taskDate.setTime(taskTimeSeconds * 1000);
+  const taskDurationMinutes = task.duration;
```
jamesnam-google, line 150: (Optional) I think task.duration should be task.durationMinutes instead. Feel free to add TODO comment if you want to make that change later.

jamesnam-google (approved):

portfolio/src/main/webapp/request.js

portfolio/src/main/webapp/request.js

portfolio/src/main/webapp/request.js

portfolio/src/main/webapp/sheetsApi.js
talvarez0507: I will leave a TODO.

PR #26, opened 2020-07-17
Array List implementation of TimeRangeGroup interface

hollyyuqizheng:


talvarez0507 (approved):

talvarez0507 (approved):

raulcruise (approved):

portfolio/src/main/java/com/google/sps/data/TimeRange.java
```diff
@@ -30,6 +30,9 @@
   public static final Comparator<TimeRange> sortByTimeRangeStartTimeAscending =
       Comparator.comparing(TimeRange::start);
 
+  public static final Comparator<TimeRange> sortByTimeRangeDurationAscending =
```
rgcase: nit: Can you rename these constants to use CAPITAL_SNAKE_CASE?

portfolio/src/main/java/com/google/sps/data/TimeRangeGroup.java
```diff
@@ -27,4 +28,11 @@
    * @return the newly modified collection of time ranges.
    */
   public Collection<TimeRange> deleteTimeRange(TimeRange timeRangeToDelete);
+
+  /**
+   * Sorts the collection of time ranges based on given comparator and a boolean value that denotes
+   * if the sorting is in ascending order. The comparator and the boolean value are passed to the
+   * constructor of the class.
+   */
+  public void sortTimeRanges(Comparator<TimeRange> comparator, boolean ascending);
```
rgcase: I don't think you need both a comparator and the `ascending` boolean. If the caller already has a `Comparator` they can reverse it themselves before passing it in.

portfolio/src/main/java/com/google/sps/data/TimeRangeGroup.java
```diff
@@ -27,4 +28,11 @@
    * @return the newly modified collection of time ranges.
    */
   public Collection<TimeRange> deleteTimeRange(TimeRange timeRangeToDelete);
+
+  /**
+   * Sorts the collection of time ranges based on given comparator and a boolean value that denotes
+   * if the sorting is in ascending order. The comparator and the boolean value are passed to the
```
rgcase: > The comparator and the boolean value are passed to the constructor of the class.

This sounds pretty dangerous to me. In the `TimeRangeGroupArrayList` below it looks like you let the caller pass in these things in the constructor, so in what scenarios would we need to change the sort order for an already existing `TimeRangeGroup`? I think we should try to treat the sort order as a fundamental immutable property of a `TimeRangeGroup`, because otherwise it'll be really hard to keep track of which sort order it's using at any time.

If you *do* need to change the sort order, I think it would be better to just create a whole new `TimeRangeGroup` with the new sort order.

portfolio/src/main/java/com/google/sps/data/TimeRangeGroupArrayList.java
```diff
@@ -0,0 +1,121 @@
...
+  public List<TimeRange> allTimeRanges;
+  private Comparator<TimeRange> comparator;
+  private boolean ascending;
+
+  /**
+   * Adds all the input time ranges to the list of all time ranges. Also sorts the list of all
+   * ranges in the constructor.
+   */
+  public TimeRangeGroupArrayList(
+      Collection<TimeRange> timeRanges, Comparator<TimeRange> comparator, boolean ascending) {
```
rgcase: Here as well, I don't think we need the `ascending` option.

portfolio/src/main/java/com/google/sps/data/TimeRangeGroupArrayList.java
```diff
@@ -0,0 +1,121 @@
...
+  /**
+   * Adds a new time range to the list. This new range must be disjoint from all existing ranges.
+   * TODO(hollyyuqizheng): Depending on how the algorithms are implemented, we might need to add the
+   * merging logic in this method. For example, if a time range is deleted but then later added back
+   * to the list for some reason, this list will potentially need to be merged.
+   */
+  public void addTimeRange(TimeRange timeRange) {
+    for (TimeRange currentRange : allTimeRanges) {
+      if (currentRange.overlaps(timeRange)) {
+        throw new IllegalArgumentException(
```
rgcase: The throwing of this exception seems a little odd to me, and I'm having a hard time seeing why we'd want to do this. Is there any way for the caller to know that they're about to trigger this? The `hasTimeRange(...)` method below checks if the given range is completely contained by some `TimeRange`, but doesn't tell you about overlaps. Is the caller meant to do this merging themselves somehow?

portfolio/src/main/java/com/google/sps/data/TimeRangeGroupArrayList.java
```diff
@@ -0,0 +1,121 @@
...
+        sortTimeRanges(comparator, ascending);
+        return allTimeRanges;
+      }
+    }
+
+    // If at this point the method hasn't returned yet, it means none of the existing
+    // time ranges actually contains the range to delete.
+    // Throw an invalid input exception here.
+    throw new IllegalArgumentException(
+        "The time range to delete must be contained by an existing time range");
```
rgcase: I'm curious here as well why we can only call this with `TimeRange`s that are completely contained in one of the ones already in the list? If the list contains `[3:00, 4:00]` why not allow the user to delete `[3:30-4:30]` and leave behind just `[3:00, 3:30]`? If the caller is in this situation, what are they supposed to do?

rgcase (changes requested):

portfolio/src/main/java/com/google/sps/data/TimeRangeGroupArrayList.java
```diff
@@ -0,0 +1,121 @@
...
+
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.List;
+
+/** Models an implementation of the TimeRangeGroup model using ArrayList. */
+public class TimeRangeGroupArrayList implements TimeRangeGroup {
```
jamesnam-google: I would suggest that we rename this class to ArrayListTimeRangeGroup. For example, ArrayList implements List, HashSet implements Set, etc.

portfolio/src/main/java/com/google/sps/data/TimeRange.java
hollyyuqizheng: Updated.

portfolio/src/main/java/com/google/sps/data/TimeRangeGroupArrayList.java
jamesnam-google: Also I don't think the comparator param is necessary as well. I think the constructor can just take in the collection.

Rationale:
1. Caller can sort the collection as they see fit.
2. If you think this class should perform sorting, I think it should do so using some reasonable sorting approach, which does not need to be specified by the caller.

portfolio/src/main/java/com/google/sps/data/TimeRangeGroupArrayList.java
jamesnam-google: I think ideally this class handles the overlap by merging the time range. For example, if 1pm-2pm was an existing time range, and this method was called with 1:30pm-2:30pm, then maybe this class can merge them and store 1pm-2:30pm.

But I think that merging logic does not need to happen in this CL. So maybe for now you can require that this method is called with a time range that does not overlap with any existing time range. Maybe you can throw UnsupportedOperationException and add a TODO comment.

portfolio/src/main/java/com/google/sps/data/TimeRangeGroupArrayList.java
```diff
@@ -0,0 +1,121 @@
...
+        throw new IllegalArgumentException(
+            "New time range to add cannot overlap with any existing time ranges");
+      }
+    }
+    allTimeRanges.add(timeRange);
+  }
+
+  /** Returns the array list of all time ranges. */
+  public List<TimeRange> getAllTimeRanges() {
+    sortTimeRanges(comparator, ascending);
```
jamesnam-google: I think in terms of sorting, you can sort when you perform set operations. Then in the get operation, you don't have to sort again.

And I think it's reasonable for this class to internally perform sorting rather than having the caller specify how to sort. I think sorting based on time range start time is reasonable.

portfolio/src/main/java/com/google/sps/data/TimeRange.java
jamesnam-google: I see that this comparator is not used anywhere. And I can't think of a scenario when we will use it. I think we probably want to always sort the time range by start time. Can we remove this variable until we come across a case where we should use it?

portfolio/src/main/java/com/google/sps/data/TimeRangeGroup.java

portfolio/src/main/java/com/google/sps/data/TimeRangeGroupArrayList.java
hollyyuqizheng: Updated.

portfolio/src/main/java/com/google/sps/data/TimeRangeGroupArrayList.java
jamesnam-google: The time range group is meant to track free time ranges. So I think conceptually it does not make sense for the user to mark 3:30pm-4:30pm as busy, when the user was free from 3-4. So I think I'm okay with throwing an exception here.

portfolio/src/main/java/com/google/sps/data/TimeRangeGroupArrayList.java
```diff
@@ -0,0 +1,121 @@
...
+
+        // Construct one or two new time ranges after the deletion.
+        if (currentRangeStart.isBefore(toDeleteRangeStart)) {
+          allTimeRanges.add(TimeRange.fromStartEnd(currentRangeStart, toDeleteRangeStart));
+        }
+
+        if (currentRangeEnd.isAfter(toDeleteRangeEnd)) {
+          allTimeRanges.add(TimeRange.fromStartEnd(toDeleteRangeEnd, currentRangeEnd));
+        }
+        // Collections.sort(allTimeRanges, TimeRange.sortByTimeRangeStartTimeAscending);
```
jamesnam-google: Should we remove this line?

portfolio/src/main/java/com/google/sps/data/TimeRangeGroupArrayList.java
```diff
@@ -0,0 +1,121 @@
...
+        sortTimeRanges(comparator, ascending);
```
jamesnam-google: I think we may not need to sort here, if allTimeRanges was already sorted.

portfolio/src/test/java/com/google/sps/TimeRangeGroupArrayListTest.java
```diff
@@ -0,0 +1,166 @@
...
+    Instant timeRangeTwoStart = timeRangeOneEnd.plusSeconds(500);
+    Instant timeRangeTwoEnd = timeRangeTwoStart.plusSeconds(1000);
+
+    TimeRange timeRangeOne = TimeRange.fromStartEnd(timeRangeOneStart, timeRangeOneEnd);
+    TimeRange timeRangeTwo = TimeRange.fromStartEnd(timeRangeTwoStart, timeRangeTwoEnd);
+
+    List<TimeRange> timeRanges = Arrays.asList(timeRangeOne, timeRangeTwo);
+    TimeRangeGroupArrayList timeRangeGroup =
+        new TimeRangeGroupArrayList(
+            timeRanges, TimeRange.sortByTimeRangeStartTimeAscending, /* ascending= */ true);
```
jamesnam-google: As I mentioned, I think sorting should be done internally by TimeRangeGroup rather than having the caller specify the sorting approach like this. For example, hash set uses a tree structure internally to keep track of the entries, but the caller doesn't need to specify how to organize the tree. So I think you shouldn't have to pass the comparator and boolean in the constructor.

jamesnam-google (changes requested):

portfolio/src/main/java/com/google/sps/data/TimeRangeGroup.java
jamesnam-google: As I mentioned below, I don't think the TimeRangeGroup interface should expose sorting to the callers. I think it should be done internally. The caller can always sort the collection that they retrieve from TimeRangeGroup the way they see fit.

portfolio/src/main/java/com/google/sps/data/TimeRangeGroupArrayList.java
```diff
@@ -0,0 +1,121 @@
...
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.List;
+
+/** Models an implementation of the TimeRangeGroup model using ArrayList. */
+public class TimeRangeGroupArrayList implements TimeRangeGroup {
+
+  public List<TimeRange> allTimeRanges;
```
jamesnam-google: Consider adding some comment that indicates that at each moment this variable is already sorted.

jamesnam-google (changes requested):

portfolio/src/main/java/com/google/sps/data/TimeRangeGroupArrayList.java
```diff
@@ -0,0 +1,121 @@
...
+      if (currentRange.overlaps(timeRange)) {
+        throw new IllegalArgumentException(
+            "New time range to add cannot overlap with any existing time ranges");
+      }
+    }
+    allTimeRanges.add(timeRange);
+  }
+
+  /** Returns the array list of all time ranges. */
+  public List<TimeRange> getAllTimeRanges() {
```
jamesnam-google: I think one issue with getAllTimeRanges is that the caller can modify the result, which can affect the internal state of ArrayListTimeRangeGroup. I think this structure is similar to HashSet in the method signatures, and you can consider exposing an iterator like hash set, rather than getAllTimeRanges. That way, callers can use the iterator to iterate over all the time ranges.

Alternatively, you can update this method so that it returns an ImmutableList rather than List.

Reference:
https://docs.oracle.com/javase/7/docs/api/java/util/HashSet.html

jamesnam-google (changes requested):

portfolio/src/main/java/com/google/sps/data/TimeRangeGroupArrayList.java
rgcase: Hmmm but the class has an `addTimeRange` method so it will have to be able to sort by itself, won't it? Also I don't think it makes sense for this class to try to choose a sort order when the caller has already sorted the collection on the way in. I think it makes sense to keep the `Comparator` as part of this class.

portfolio/src/main/java/com/google/sps/data/TimeRangeGroupArrayList.java
rgcase: I agree that the class should handle the merging, but I'm not sure if it's a good idea to leave it to a different CL.

> So maybe for now you can require that this method is called with a time range that does not overlap with any existing time range.

I don't think there's a good way for the caller to know whether or not this is true is there? I guess they can call `getAllTimeRanges()` and check them all first, but that seems pretty awkward.

portfolio/src/main/java/com/google/sps/data/TimeRangeGroupArrayList.java
rgcase: Maybe we should rename this to `FreeTimeRangeGroup` then? But I don't see why we wouldn't write it more generally. Also the same situation as above applies here, where the caller doesn't really have any way to know before calling this method whether or not the input will be valid.

Also, I don't see why deleting [3:30, 4:30] is any less valid than deleting [3:30, 4:00] would be? And what if the group contained both [3:00, 4:00] and [4:00-5:00], should it be valid to delete [3:30, 4:30] then?

portfolio/src/main/java/com/google/sps/data/TimeRangeGroupArrayList.java
rgcase: I think it would also be fine if you *do* want to just choose a sort order in here and stick with it, but in that case there should be no guarantee that the sort order of the list as it's passed in is preserved at all.

portfolio/src/main/java/com/google/sps/data/TimeRangeGroupArrayList.java
rgcase: nit: I think you can make this `final` as well.

portfolio/src/main/java/com/google/sps/data/TimeRange.java
hollyyuqizheng: Updated. In the longest task first algorithm that I am also working on, I actually used a comparator that sorts time ranges by duration first and then start time, so I updated the constant variables for the comparator accordingly here.

portfolio/src/main/java/com/google/sps/data/TimeRangeGroupArrayList.java
hollyyuqizheng: Updated the class name to ArrayListTimeRangeGroup

portfolio/src/main/java/com/google/sps/data/TimeRangeGroupArrayList.java
hollyyuqizheng: Updated with comment.

portfolio/src/main/java/com/google/sps/data/TimeRangeGroupArrayList.java
hollyyuqizheng: Updated.

portfolio/src/main/java/com/google/sps/data/TimeRangeGroupArrayList.java
hollyyuqizheng: Updated.

portfolio/src/main/java/com/google/sps/data/TimeRangeGroupArrayList.java
hollyyuqizheng: Updated by returning an iterator.

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
```diff
@@ -0,0 +1,169 @@
...
+        (range) -> {
+          addTimeRange(range);
+        });
+  }
+
+  /**
+   * Adds a new time range to the list. If the time range to add overlaps with any existing time
+   * range, the overlapping time ranges will be merged.
+   */
+  public void addTimeRange(TimeRange timeRange) {
```
jamesnam-google, line 32: I think there is an issue with this approach. Consider the case where allTimeRanges look like:
1. 1pm-2pm
2. 5pm-6pm
3. 3pm-4pm

And suppose timeRange is 1:30pm-3:30pm.

The desired end state of the internal list is:
1. 1pm-4pm
2. 5pm-6pm

But I think instead this approach will yield:
1. 1pm-3:30pm
2. 5pm-6pm
3. 3pm-4pm

I strongly recommend that you keep the internal list sorted. I think that should allow you to implement the algorithm with fewer boundary conditions and complex cases.

jamesnam-google (changes requested):

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
hollyyuqizheng: I modified the `testAddCompletelyOverlappingRange` test method for this test case.

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
```diff
@@ -0,0 +1,171 @@
...
+      } else {
+        // This is the case that lastExaminedTimeRange and the current time range do not overlap.
+        // If the current range from the original list ends after the last examined time range,
+        // add the time range pointed to by lastExaminedTimeRange to the new list of all ranges, and
+        // change the lastExaminedTimeRange to point to the current range.
+        if (currentRange.end().isAfter(lastExaminedTimeRange.end())) {
+          newTimeRanges.add(lastExaminedTimeRange);
+          lastExaminedTimeRange = currentRange;
+        } else {
+          // If lastExamiendTimeRange is later than the current range,
```
talvarez0507: nit: examined

portfolio/src/main/java/com/google/sps/data/TimeRange.java
```diff
@@ -26,10 +26,14 @@
   private final Instant start;
   private final Duration duration;
 
-  // Comparator for sorting time ranges by start time
-  public static final Comparator<TimeRange> sortByTimeRangeStartTimeAscending =
+  // Comparator for sorting time ranges  by start time ascending
```
talvarez0507, line 29: nit: extra space

portfolio/src/test/java/com/google/sps/ArrayListTimeRangeGroupTest.java
```diff
@@ -0,0 +1,522 @@
...
+    List<TimeRange> actual = new ArrayList();
+    while (actualIterator.hasNext()) {
+      actual.add(actualIterator.next());
+    }
+
+    Collections.sort(actual, TimeRange.SORT_BY_TIME_RANGE_DURATION_ASCENDING_THEN_START_TIME);
+    Assert.assertEquals(expected, actual);
+  }
+
+  /** Tests for adding a time range that overlaps with all other existing ones. */
```
talvarez0507: nit:  "with all other existing" should be "with several other existing"

talvarez0507 (approved):

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
```diff
@@ -0,0 +1,171 @@
...
+    // this is the first time we add anything to the list,
+    // so simply add the new time range to the list and return.
+    if (allTimeRanges.isEmpty()) {
+      allTimeRanges.add(timeRange);
+      return;
+    }
+
+    List<TimeRange> newTimeRanges = new ArrayList<TimeRange>();
+
+    // This variable represents the latest time range previously exmained.
```
jamesnam-google: exmained => examined

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
```diff
@@ -0,0 +1,171 @@
...
+    // Initially, this variable points to the time range we want to add.
+    // As the for loop iterates through all the current time ranges already existing
+    // in allTimeRanges, this lastExaminedTimeRange variable is the time range with which
+    // any existing time range merges with, if necessary.
+    TimeRange lastExaminedTimeRange = timeRange;
+
+    for (int i = 0; i < allTimeRanges.size(); i++) {
+      TimeRange currentRange = allTimeRanges.get(i);
+
+      // If the current range is complete contained by the lastExaminedTimeRange,
```
jamesnam-google: complete contained => completely contained

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
```diff
@@ -0,0 +1,171 @@
...
+      // Change the lastExaminedTimeRange pointer to the new time range created
+      // after the merging.
+      if (lastExaminedTimeRange.overlaps(currentRange)) {
+        lastExaminedTimeRange = mergeTwoTimeRanges(currentRange, lastExaminedTimeRange);
+      } else {
+        // This is the case that lastExaminedTimeRange and the current time range do not overlap.
+        // If the current range from the original list ends after the last examined time range,
+        // add the time range pointed to by lastExaminedTimeRange to the new list of all ranges, and
+        // change the lastExaminedTimeRange to point to the current range.
+        if (currentRange.end().isAfter(lastExaminedTimeRange.end())) {
```
jamesnam-google: We can make this else if

ex:
if (lastExamined...) {
  ...
} else if (currentRange...) {
  ...
} else {
  ...
}

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
```diff
@@ -0,0 +1,171 @@
...
+      // If current time range is the last element in the allTimeRanges list,
+      // add the last examined time range to the new list.
+      if (i == allTimeRanges.size() - 1) {
+        newTimeRanges.add(lastExaminedTimeRange);
+      }
+    }
+
+    // Finally, set the global variable allTimeRanges to this newly built list of time ranges.
+    allTimeRanges = newTimeRanges;
+    Collections.sort(allTimeRanges, TimeRange.SORT_BY_TIME_RANGE_START_TIME_ASCENDING);
```
jamesnam-google: I don't think this call is necessary, if we ensure that when this method is invoked the internal list is already sorted.

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
```diff
@@ -0,0 +1,171 @@
...
+          addTimeRange(range);
+        });
+    Collections.sort(allTimeRanges, TimeRange.SORT_BY_TIME_RANGE_START_TIME_ASCENDING);
+  }
+
+  /**
+   * Adds a new time range to the list. If the time range to add overlaps with any existing time
+   * range, the overlapping time ranges will be merged.
+   */
+  public void addTimeRange(TimeRange timeRange) {
```
jamesnam-google, line 32: Currently, addTimeRange is O(n). But I think it can be made O(log n) by using binary search. So I think the pod can consider making that optimization sometime.

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
```diff
@@ -0,0 +1,171 @@
...
+      newTimeRangeEnd = b.end();
+    } else {
+      newTimeRangeEnd = a.end();
+    }
+
+    return TimeRange.fromStartEnd(newTimeRangeStart, newTimeRangeEnd);
+  }
+
+  /** Returns an iterator for the list of all time ranges. */
+  public Iterator<TimeRange> getAllTimeRanges() {
```
jamesnam-google: Consider renaming the method to iterator

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
```diff
@@ -0,0 +1,171 @@
...
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.List;
+
+/** Models an implementation of the TimeRangeGroup model using ArrayList. */
+public class ArrayListTimeRangeGroup implements TimeRangeGroup {
+
+  // This list of all the time ranges will be sorted by start time ascending
+  // in add and delete methods.
```
jamesnam-google: Consider clarifying in the comment that the time ranges in allTimeRanges are always pair-wise disjoint.

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
```diff
@@ -0,0 +1,171 @@
...
+   * Adds all the input time ranges to the list of all time ranges. Also sorts the list of all
+   * ranges in the constructor.
+   */
+  public ArrayListTimeRangeGroup(Collection<TimeRange> timeRanges) {
+    allTimeRanges = new ArrayList<TimeRange>();
+    timeRanges.forEach(
+        (range) -> {
+          addTimeRange(range);
+        });
+    Collections.sort(allTimeRanges, TimeRange.SORT_BY_TIME_RANGE_START_TIME_ASCENDING);
```
jamesnam-google: I'm not sure whether this sort call is necessary. I think even without it allTimeRanges will be already sorted.

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
```diff
@@ -0,0 +1,171 @@
...
+      }
+    }
+
+    // Finally, set the global variable allTimeRanges to this newly built list of time ranges.
+    allTimeRanges = newTimeRanges;
+    Collections.sort(allTimeRanges, TimeRange.SORT_BY_TIME_RANGE_START_TIME_ASCENDING);
+  }
+
+  /** Helper method for merging two time ranges. This method is public so that it can be tested. */
+  public static TimeRange mergeTwoTimeRanges(TimeRange a, TimeRange b) {
```
jamesnam-google: Can we make this method package-private?

jamesnam-google: Consider throwing exception if there is no overlap between the two time ranges.

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
```diff
@@ -0,0 +1,171 @@
...
+  /** Returns an iterator for the list of all time ranges. */
+  public Iterator<TimeRange> getAllTimeRanges() {
+    return allTimeRanges.iterator();
+  }
+
+  /**
+   * Checks if a time range exists in the collection. For example, if [3:00 - 4:00] is in the
+   * collection, [3:00 - 3:30] is considered to exist as a time range in the collection.
+   */
+  public boolean hasTimeRange(TimeRange timeRangeToCheck) {
```
jamesnam-google, line 133: It'd be nice to use binary search for a method like this in the future, if you can.

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
```diff
@@ -0,0 +1,171 @@
...
+        if (currentRangeEnd.isAfter(toDeleteRangeEnd)) {
+          newTimeRanges.add(TimeRange.fromStartEnd(toDeleteRangeEnd, currentRangeEnd));
+        }
+      } else {
+        newTimeRanges.add(currentRange);
+      }
+    }
+
+    allTimeRanges = newTimeRanges;
+    Collections.sort(allTimeRanges, TimeRange.SORT_BY_TIME_RANGE_START_TIME_ASCENDING);
```
jamesnam-google: I am not sure whether you need this sort call.

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
```diff
@@ -0,0 +1,171 @@
...
+   * <p>For example, if the list contains [3:00 - 4:00] and [5:00 - 6:00], deleting [3:15 - 3:30]
+   * will result in [3 - 3:15] and [3:30 - 4] as new time ranges.
+   *
+   * <p>Another example for deleting overlapping time ranges: if [3 - 4] and [5 - 6] are in the
+   * original list, deleting [3:30 - 5:30] will result in two new ranges: [3 - 3:30] and [5:30 - 6].
+   */
+  public void deleteTimeRange(TimeRange timeRangeToDelete) {
+    List<TimeRange> newTimeRanges = new ArrayList<TimeRange>();
+
+    for (TimeRange currentRange : allTimeRanges) {
```
jamesnam-google, line 166: What about the case where currentRange is completely covered by timeRangeToDelete?

jamesnam-google (changes requested):

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
hollyyuqizheng: Updated.

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
hollyyuqizheng: Updated.

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
hollyyuqizheng: Updated.

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
hollyyuqizheng: Updated.

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
hollyyuqizheng: Updated.

hollyyuqizheng: Updated.

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
hollyyuqizheng: Updated.

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
hollyyuqizheng: I think in that case, none of the two if statements inside the `if overlaps` gets executed, so the current time range that is covered by the range to delete is essentially ignored at this iteration of the for loop. I added a test case for this scenario.

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
hollyyuqizheng: Updated.

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
```diff
@@ -0,0 +1,187 @@
...
+  // pair-wise disjoint at any moment.
+  public List<TimeRange> allTimeRanges;
+
+  /**
+   * Adds all the input time ranges to the list of all time ranges. Also sorts the list of all
+   * ranges in the constructor.
+   */
+  public ArrayListTimeRangeGroup(Collection<TimeRange> timeRanges) {
+    allTimeRanges = new ArrayList<TimeRange>();
+    timeRanges.forEach(
```
rgcase, line 22: nit: I think this can just be `timeRanges.forEach(addTimeRange)`.

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
```diff
@@ -0,0 +1,187 @@
...
+import java.util.Iterator;
+import java.util.List;
+
+/** Models an implementation of the TimeRangeGroup model using ArrayList. */
+public class ArrayListTimeRangeGroup implements TimeRangeGroup {
+
+  // This list of all the time ranges will be sorted by start time ascending
+  // in add and delete methods. The time ranges stored in this list are
+  // pair-wise disjoint at any moment.
+  public List<TimeRange> allTimeRanges;
```
rgcase: nit: This should be private, so callers need to go through your API to get the result which gives you a chance to make an iterator or defensive copy.

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
```diff
@@ -0,0 +1,187 @@
...
+      newTimeRangeEnd = b.end();
+    } else {
+      newTimeRangeEnd = a.end();
+    }
+
+    return TimeRange.fromStartEnd(newTimeRangeStart, newTimeRangeEnd);
+  }
+
+  /** Returns an iterator for the list of all time ranges. */
+  public Iterator<TimeRange> getAllTimeRangesIterator() {
```
rgcase: I *think* if you call this method just `iterator()`, then it'll automatically be possible to use it in foreach syntax like:
```
for (TimeRange timeRange : timeRangeGroup) {
  ...
}
```

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
jamesnam-google: +1

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
```diff
@@ -0,0 +1,187 @@
...
+  // This list of all the time ranges will be sorted by start time ascending
+  // in add and delete methods. The time ranges stored in this list are
+  // pair-wise disjoint at any moment.
+  public List<TimeRange> allTimeRanges;
+
+  /**
+   * Adds all the input time ranges to the list of all time ranges. Also sorts the list of all
+   * ranges in the constructor.
+   */
+  public ArrayListTimeRangeGroup(Collection<TimeRange> timeRanges) {
```
jamesnam-google: I think you might be able to take in Iterable<TimeRange> that is more general than Collection

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
```diff
@@ -0,0 +1,187 @@
...
+    // in allTimeRanges, this lastExaminedTimeRange variable is the time range with which
+    // any existing time range merges with, if necessary.
+    TimeRange lastExaminedTimeRange = timeRange;
+
+    for (int i = 0; i < allTimeRanges.size(); i++) {
+      TimeRange currentRange = allTimeRanges.get(i);
+
+      // If the current range is completely contained by the lastExaminedTimeRange,
+      // no merging or adding needs to happen.
+      if (lastExaminedTimeRange.contains(currentRange)) {
```
jamesnam-google, line 55: I think there is an issue.

Suppose that the internal list had one item, which is 1pm-2pm. add method is called with 12pm-3pm. The desired end state for the internal list is 12pm-3pm. But I think it stays as 1pm-2pm.

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
```diff
@@ -0,0 +1,187 @@
...
+  public boolean hasTimeRange(TimeRange timeRangeToCheck) {
+    int end = allTimeRanges.size() - 1;
+    int start = 0;
+    int middle;
+
+    while (start < end) {
+      middle = (start + end) / 2;
+      TimeRange middleRange = allTimeRanges.get(middle);
+      if (middleRange.start().isAfter(timeRangeToCheck.start())) {
+        end = middle;
```
jamesnam-google: I think we can assign end = middle - 1, because middleRange cannot contain timeRangeToCheck.

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
jamesnam-google: I see. Okay thanks.

jamesnam-google (changes requested):

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
hollyyuqizheng: Updated.

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
hollyyuqizheng: Updated.

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
hollyyuqizheng: I see. Updated accordingly.

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
hollyyuqizheng: Updated.

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
hollyyuqizheng: I see. I updated this if statement to check if the current range is the last element in the original list.

raulcruise (approved):

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
```diff
@@ -0,0 +1,192 @@
...
+      newTimeRangeEnd = b.end();
+    } else {
+      newTimeRangeEnd = a.end();
+    }
+
+    return TimeRange.fromStartEnd(newTimeRangeStart, newTimeRangeEnd);
+  }
+
+  /** Returns an iterator for the list of all time ranges. */
+  // public Iterator<TimeRange> getAllTimeRangesIterator() {
```
jamesnam-google: Can we remove these comment lines?

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
```diff
@@ -0,0 +1,192 @@
...
+   * and whose end time is after the target range's end. Then the method calls contains to see if
+   * the target range is contained within this current range.
+   */
+  public boolean hasTimeRange(TimeRange timeRangeToCheck) {
+    int end = allTimeRanges.size() - 1;
+    int start = 0;
+    int middle;
+
+    while (start < end) {
+      middle = (start + end) / 2;
```
jamesnam-google: I think you can declare the middle variable here.

int middle = ..

jamesnam-google (approved):

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
hollyyuqizheng: Updated.

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
hollyyuqizheng: Updated.

rgcase (approved):

PR #27, opened 2020-07-21
Add a .equals() for CalendarEvent

talvarez0507:
This PR just adds a .equals() to the CalendarEvent class. In a recent PR I added some tests for ServletHelper.java and after pulling master I realized these tests are failing. Without the overrided .equals(), which I forgot to include in that PR, these tests are checking whether the CalendarEvents are the same object and not whether the fields are the same.

### DIFF
--- a/portfolio/src/main/java/com/google/sps/data/CalendarEvent.java
+++ b/portfolio/src/main/java/com/google/sps/data/CalendarEvent.java
@@ -48,4 +48,14 @@ public class CalendarEvent {
   public Instant getEndTime() {
     return endTime;
   }
+  @Override
+  public boolean equals(Object other) {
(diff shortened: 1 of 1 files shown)


hollyyuqizheng (approved):

raulcruise (approved):

portfolio/src/main/java/com/google/sps/data/CalendarEvent.java
```diff
@@ -48,4 +48,14 @@ public Instant getStartTime() {
   public Instant getEndTime() {
     return endTime;
   }
+  @Override
```
jamesnam-google, line 52: Line break above this line.

jamesnam-google (approved):

rgcase (approved):

PR #30, opened 2020-07-23
Update API JavaScript files 

raulcruise:
This pull request mostly consists of the authentication functions being removed from each API file since they'll all now be in apiLibrary.js. Only a couple of new functionalities were added to match the currently deployed MVP.

### DIFF
--- a/portfolio/src/main/webapp/sheetsApi.js
+++ b/portfolio/src/main/webapp/sheetsApi.js
@@ -1,91 +1,27 @@
-// Client ID and API key from the Developer Console
-const SHEETS_CLIENT_ID =
-    '499747085593-hvi6n4kdrbbfvcuo1c9a9tu9oaf62cr2.apps.googleusercontent.com';
-const SHEETS_API_KEY = '';
-
-// Array of API discovery doc URLs for APIs used by the quickstart
-const SHEETS_DISCOVERY_DOCS =
-    ['https://sheets.googleapis.com/$discovery/rest?version=v4'];
-
-// Authorization scopes required by the API; multiple scopes can be
-// included, separated by spaces.
-const SCOPE_SHEETS_READ_WRITE = 'https://www.googleapis.com/auth/spreadsheets';
-
 const URL_DOCUMENT_BEGINNING = 'https://docs.google.com/spreadsheets/d/';
-
-// Error codes that the process can throw
-const ERROR_CODE_SHEETS = {
-  POPUP_CLOSED: 'popup_closed_by_user',
-  ACCESS_DENIED: 'access_denied'
-};
-
-/**
- * On load, called to load the auth2 library and API client library.
- */
-function handleClientLoadSheets() {
-  gapi.load('client:auth2', initClientSheets);
-}
 
 /**
  * Initializes the API client library and sets up sign-in state
  * listeners.
  */
 function initClientSheets() {
-  gapi.client
-      .init({
-        apiKey: SHEETS_API_KEY,
-        clientId: SHEETS_CLIENT_ID,
-        discoveryDocs: SHEETS_DISCOVERY_DOCS,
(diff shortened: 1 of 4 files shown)


hollyyuqizheng (approved):

portfolio/src/main/webapp/sheetsApi.js
```diff
@@ -1,91 +1,27 @@
...
-}
-
-
 /**
  * Called when the sign-in state changes and updates the UI appropriately.
  */
 function handleUiSheets() {
   $('#sheets-export-button').addClass('d-none');
-  if (gapi.auth2.getAuthInstance().isSignedIn.get()) {
+  if (GoogleAuth.isSignedIn.get()) {
```
talvarez0507, line 24: This is just a question. Why did we change some GoogleAuth's to gapi.auth2... and vice versa. Was there no way of making it all GoogleAuth or gapi.auth2?

portfolio/src/main/webapp/tasksApi.js
```diff
@@ -102,6 +54,11 @@ function importTasklist(tasklistId) {
  * logged in.
  */
 function drawImportMenu() {
+  importMenuVisible = true;
+
+  const $button = $('#connect-tasks-btn')[0];
```
talvarez0507: Just a small suggestion, the variable name button seems a little vague.

talvarez0507 (approved):

portfolio/src/main/webapp/sheetsApi.js
raulcruise: Good catch, I'll change them all to GoogleAuth.

raulcruise: Done.

portfolio/src/main/webapp/tasksApi.js
raulcruise: Done.

rgcase (approved):

portfolio/src/main/webapp/tasksApi.js
```diff
@@ -143,19 +100,15 @@ function drawImportMenu() {
 }
 
 function clearImportMenu() {
+  importMenuVisible = false;
+
+  const $toggleTasksButton = $('#connect-tasks-btn')[0];
```
jamesnam-google: Is this a jquery variable? If not, I think it shouldn't start with $

portfolio/src/main/webapp/tasksApi.js
```diff
@@ -102,6 +54,11 @@ function importTasklist(tasklistId) {
  * logged in.
  */
 function drawImportMenu() {
+  importMenuVisible = true;
+
+  const $toggleTasksButton = $('#connect-tasks-btn')[0];
```
jamesnam-google: Is this a jquery variable? If not, I think it shouldn't start with $

portfolio/src/main/webapp/tasksApi.js
```diff
@@ -1,64 +1,16 @@
-var GoogleAuth;
+var importMenuVisible;
```
jamesnam-google: I think we should not keep track in Javascript like this. I think we should keep track of the state in the DOM. For example, it can be some attribute in the toggle tasks button, like data-is-import-menu-visible="true".

jamesnam-google (approved):

portfolio/src/main/webapp/tasksApi.js
raulcruise: Done.

portfolio/src/main/webapp/tasksApi.js
raulcruise: Done.

portfolio/src/main/webapp/tasksApi.js
raulcruise: Done.

PR #31, opened 2020-07-23
Move API Authentication to a separate file

raulcruise:
Currently in the master branch each API file has its own authentication function(s), however this pull requests simplifies and actually fixes our API authentication since the master branch currently breaks when using more than one API. The first authentication flow is moved to apiLibrary.js.

### DIFF
--- a/portfolio/src/main/webapp/apiLibrary.js
+++ b/portfolio/src/main/webapp/apiLibrary.js
@@ -1,0 +1,177 @@
+const CLIENT_ID =
+    '499747085593-hvi6n4kdrbbfvcuo1c9a9tu9oaf62cr2.apps.googleusercontent.com';
+
+var GoogleAuth;
+
+// Constants for discovery documents' URLs.
+const DISCOVERY_DOCS_CALENDAR =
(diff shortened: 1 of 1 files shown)


hollyyuqizheng (approved):

talvarez0507 (approved):

portfolio/src/main/webapp/apiLibrary.js
```diff
@@ -0,0 +1,177 @@
...
+ * Loads the API's client and auth2 modules.
+ * Calls the initCalendarClient function after the modules load.
+ */
+function handleClientLoad() {
+  gapi.load('client:auth2', initAuthentication);
+}
+
+function initAuthentication() {
+  fetchApiKey().then(
+      responseJson => gapi.client
```
jamesnam-google: Consider using a helper method for gapi.client.init call (it can take responseJson as a param)

jamesnam-google (approved):

rgcase (approved):

portfolio/src/main/webapp/apiLibrary.js
raulcruise: Done.

PR #32, opened 2020-07-23
Update HTML and CSS files

raulcruise:
The HTML file is updated to reflect the changes made to different function names and new functions added in the API files as well as the files for basic features. The CSS file is updated with some style changes updated in the MVP.

### DIFF
--- a/portfolio/src/main/webapp/style.css
+++ b/portfolio/src/main/webapp/style.css
@@ -20,7 +20,7 @@
   width: 100%;
 }
 
-.btn-primary {
+.btn-with-top-margin {
   margin-top: 2px;
 }
 .card {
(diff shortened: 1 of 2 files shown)


hollyyuqizheng (approved):

talvarez0507 (approved):

rgcase (approved):

portfolio/src/main/webapp/style.css
```diff
@@ -1,35 +1,46 @@
...
+
 #nav-bar {
   background-color: #dae8fe;
   width: 100%;
 }
- 
-#new-event-list-wrapper {
-  width: 80%;
+
+.btn-primary {
```
jamesnam-google: This style seems to general. I think you want to be more specific. Otherwise all primary buttons will get extra margin top.

jamesnam-google (approved):

portfolio/src/main/webapp/style.css
raulcruise: Done.

PR #33, opened 2020-07-24
Longest task first algorithm

hollyyuqizheng:
This PR contains the algorithm for scheduling the longest tasks first so that as many longer tasks get scheduled as possible. I also refactored some code from the original ShortestTaskFirstSchedulerTest.java into a util file for scheduler tests.


talvarez0507 (approved):

raulcruise (approved):

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java
```diff
@@ -0,0 +1,88 @@
...
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.Iterator;
+import java.util.List;
+
+public class LongestTaskFirstScheduler implements TaskScheduler {
+
+  // Comparator for sorting tasks by duration in ascending order
```
jamesnam-google: I see that you reverse this order below. What do you think about having the variable sort in descending order?

ex:
sortByTaskDurationDescending

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java
```diff
@@ -0,0 +1,88 @@
...
+    TimeRangeGroup availableTimesGroup = new ArrayListTimeRangeGroup(availableTimes);
+
+    List<ScheduledTask> scheduledTasks = new ArrayList<ScheduledTask>();
+
+    for (Task task : tasksList) {
+      Duration taskDuration = task.getDuration();
+
+      // Find the first free time range that is longer than the current task's duration.
+      for (TimeRange currentFreeTimeRange : availableTimes) {
+        if (currentFreeTimeRange.duration().compareTo(taskDuration) >= 0) {
```
jamesnam-google: I think it'd be simpler to continue in the opposite case.

ex:
if (currentFreeTimeRange.duration().compareTo(taskDuration) < 0) {
  continue;
}

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java
```diff
@@ -0,0 +1,88 @@
...
+    // Create a TimeRangeGroup class for the free time ranges.
+    TimeRangeGroup availableTimesGroup = new ArrayListTimeRangeGroup(availableTimes);
+
+    List<ScheduledTask> scheduledTasks = new ArrayList<ScheduledTask>();
+
+    for (Task task : tasksList) {
+      Duration taskDuration = task.getDuration();
+
+      // Find the first free time range that is longer than the current task's duration.
+      for (TimeRange currentFreeTimeRange : availableTimes) {
```
jamesnam-google: I think that availableTimes should be just used to construct availableTimesGroup. I think afterward, you shouldn't use it anymore.

I would suggest that you create a helper method such as,

private static Optional<TimeRange> findTimeForTask(Task task, TimeRangeGroup availableTimesGroup) {
  ...
}

Note that it can return optional empty if there is no time for the task. And I think you'll want to handle those cases as well.

Then the for loop here will look something like,

for (Task task : tasksList) {
  Optional<TimeRange> timeRangeForTask = findTimeForTask(task, availableTimesGroup);
  if (!timeRangeForTask.isPresent()) {
    // Handle, then continue.
  }
  // Create and append scheduled tasks.
  availableTimesGroup.deleteTimeRange(timeRangeForTask.get());
}

As you can see, I don't think you need the iterator and while loop logic you have below. And I don't think you need to sort availableTimes either.

portfolio/src/main/java/com/google/sps/data/TimeRange.java
```diff
@@ -26,10 +26,14 @@
   private final Instant start;
   private final Duration duration;
 
-  // Comparator for sorting time ranges by start time
-  public static final Comparator<TimeRange> sortByTimeRangeStartTimeAscending =
+  // Comparator for sorting time ranges  by start time ascending
```
jamesnam-google: Unnecessary space

portfolio/src/main/java/com/google/sps/data/TimeRange.java
```diff
@@ -26,10 +26,14 @@
...
+  public static final Comparator<TimeRange> SORT_BY_TIME_RANGE_START_TIME_ASCENDING =
       Comparator.comparing(TimeRange::start);
 
+  // Comparator for sorting time ranges by duration ascending and then by start time ascending
+  public static final Comparator<TimeRange> SORT_BY_TIME_RANGE_DURATION_ASCENDING_THEN_START_TIME =
```
jamesnam-google, line 29: I'm not sure why this comparator is necessary. I think TimeRangeGroup already keeps the internal list sorted by time window start time. Can we try not to sort again by time window duration?

portfolio/src/main/java/com/google/sps/data/TimeRangeGroup.java
```diff
@@ -22,9 +19,13 @@
...
-   * @return the newly modified collection of time ranges.
+   * will replace the original [3:00 - 4:00]. Another example for deleting overlapping time ranges:
+   * if [3 - 4] and [5 - 6] are in the original list, deleting [3:30 - 5:30] will result in two new
+   * ranges: [3 - 3:30] and [5:30 - 6].
    */
-  public Collection<TimeRange> deleteTimeRange(TimeRange timeRangeToDelete);
+  public void deleteTimeRange(TimeRange timeRangeToDelete);
+
+  /** Returns an iterator for the collection of all time ranges. */
+  // public Iterator<TimeRange> getAllTimeRangesIterator();
```
jamesnam-google: Can we remove this comment?

jamesnam-google (changes requested):

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
```diff
@@ -0,0 +1,192 @@
...
+      newTimeRangeEnd = b.end();
+    } else {
+      newTimeRangeEnd = a.end();
+    }
+
+    return TimeRange.fromStartEnd(newTimeRangeStart, newTimeRangeEnd);
+  }
+
+  /** Returns an iterator for the list of all time ranges. */
+  // public Iterator<TimeRange> getAllTimeRangesIterator() {
```
jamesnam-google: Can we remove these comments?

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java

portfolio/src/main/java/com/google/sps/data/TimeRange.java
hollyyuqizheng: Updated.

portfolio/src/main/java/com/google/sps/data/TimeRange.java
hollyyuqizheng: I see. Yes, it's not necessary anymore. It was originally used in testing.

portfolio/src/main/java/com/google/sps/data/TimeRangeGroup.java

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java
hollyyuqizheng: I see, updated accordingly. Now the tasks are scheduled to the first time range that is long enough for the task.

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java

portfolio/src/main/java/com/google/sps/data/TimeRange.java
hollyyuqizheng: After I changed the for loop logic in schedule method, this comparator is not necessary anymore, but this comparator is currently used in tests for ArrayListTimeRangeGroup, so I kept this comparator here.

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java
```diff
@@ -0,0 +1,104 @@
...
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Optional;
+
+public class LongestTaskFirstScheduler implements TaskScheduler {
```
jamesnam-google, line 14: Consider adding some class comment

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java
```diff
@@ -0,0 +1,104 @@
...
+
+  // Comparator for sorting tasks by duration in descending order and then by task name
+  // alphabetically.
+  public static final Comparator<Task> sortByTaskDurationDescendingThenName =
```
jamesnam-google: You may want to use the task priority as the second sort param, rather than name

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java
```diff
@@ -0,0 +1,104 @@
...
+      Comparator.comparing(Task::getDuration).reversed().thenComparing(Task::getName);
+
+  /**
+   * Schedules the tasks so that the longest tasks are scheduled to the first possible free time
+   * range of the day. This approach tries to prioritize long tasks so that they are scheduled.
+   */
+  public Collection<ScheduledTask> schedule(
```
jamesnam-google, line 25: Consider clarifying in the method comment that tasks can be modified by this method. Or you can update the method so that you don't modify tasks param, and just copy it into a new list. I think the latter might be a good idea.

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java
```diff
@@ -0,0 +1,104 @@
...
+        new CalendarEventsGroup(eventsList, workHoursStartTime, workHoursEndTime);
+    // Create a TimeRangeGroup class for the free time ranges.
+    List<TimeRange> availableTimes = calendarEventsGroup.getFreeTimeRanges();
+    TimeRangeGroup availableTimesGroup = new ArrayListTimeRangeGroup(availableTimes);
+
+    List<ScheduledTask> scheduledTasks = new ArrayList<ScheduledTask>();
+
+    for (Task task : tasksList) {
+      Duration taskDuration = task.getDuration();
+      Optional timeRangeForTask = findTimeForTask(task, availableTimesGroup);
```
jamesnam-google: Can we make the type Optional<TimeRange> ? Sorry I think I might have forgotten to mention this earlier.

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java
```diff
@@ -0,0 +1,104 @@
...
+
+      if (!timeRangeForTask.isPresent()) {
+        continue;
+      }
+
+      // Create and append scheduled tasks.
+      TimeRange scheduledTimeRange = (TimeRange) timeRangeForTask.get();
```
jamesnam-google: Then you don't have to do the type casting with TimeRange.

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java
```diff
@@ -0,0 +1,104 @@
...
+    return SchedulingAlgorithmType.LONGEST_TASK_FIRST;
+  }
+
+  /**
+   * Tries to find a free time range for a task.
+   *
+   * @return an Optional that contains the time range scheduled for this task. This Optional object
+   *     is empty if the task cannot be scheduled.
+   */
+  private static Optional findTimeForTask(Task task, TimeRangeGroup availableTimesGroup) {
```
jamesnam-google: Let's return Optional<TimeRange>

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java
```diff
@@ -0,0 +1,104 @@
...
+    List<TimeRange> availableTimes = new ArrayList<TimeRange>();
+    Iterator<TimeRange> availableTimesIterator = availableTimesGroup.iterator();
+    while (availableTimesIterator.hasNext()) {
+      availableTimes.add(availableTimesIterator.next());
+    }
+
+    Duration taskDuration = task.getDuration();
+    Optional scheduledTimeRangeOptional = Optional.empty();
```
jamesnam-google: Optional<TimeRange>

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java
```diff
@@ -0,0 +1,104 @@
...
+
+    // Find the first free time range that is longer than the current task's duration.
+    for (TimeRange currentFreeTimeRange : availableTimes) {
```
jamesnam-google: Consider performing this for loop logic in the while loop above. That way, we can return early without exhaustively going through availableTimesIterator.

jamesnam-google (approved):

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java
hollyyuqizheng: Updated.

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
hollyyuqizheng: Updated.

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java
hollyyuqizheng: I think the initialization of `tasksList` is equivalent to copying the tasks into a new list?

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java
hollyyuqizheng: I updated the type to `Optional<TimeRange>`.

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java
hollyyuqizheng: I updated it to return `Optional<TimeRange>`.

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java
hollyyuqizheng: Updated. I also added a `compareTo` method in TaskPriority.java

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
```diff
@@ -0,0 +1,189 @@
...
+  }
+
+  /**
+   * Checks if a time range exists in the collection. For example, if [3:00 - 4:00] is in the
+   * collection, [3:00 - 3:30] is considered to exist as a time range in the collection. This method
+   * uses binary search to find the time ranges whose start time is before the target range's start
+   * and whose end time is after the target range's end. Then the method calls contains to see if
+   * the target range is contained within this current range.
+   */
+  public boolean hasTimeRange(TimeRange timeRangeToCheck) {
```
rgcase, line 133: I think this will crash with an `IndexOutOfBoundsException` if the group is empty since you'll end up with `start == end == 0`, skip the loop then call `allTimeRanges.get(0)`?

Can you special case the empty set to return `false` and then add a test?

rgcase (approved):

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
hollyyuqizheng: I see. Updated.

PR #34, opened 2020-07-27
Linked List implementation of TimeRangeGroup interface

talvarez0507:
This is essentially the ArrayList implementation but everything was changed to linked lists, tests included. I think every time complexity is strictly worse in this implementation while memory is the same. Improving some of them some of them is complicated as we discussed so I'm not sure which ones I should improve.

### DIFF
--- a/portfolio/src/main/java/com/google/sps/data/LinkedListTimeRangeGroup.java
+++ b/portfolio/src/main/java/com/google/sps/data/LinkedListTimeRangeGroup.java
@@ -1,0 +1,180 @@
+package com.google.sps.data;
+
+import java.time.Instant;
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.LinkedList;
+import java.util.List;
+
+/**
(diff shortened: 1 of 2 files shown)


portfolio/src/main/java/com/google/sps/data/LinkedListTimeRangeGroup.java
```diff
@@ -0,0 +1,180 @@
...
+
+    // This variable represents the latest time range previously exmamined.
+    // Initially, this variable points to the time range we want to add.
+    // As the for loop iterates through all the current time ranges already existing
+    // in allTimeRanges, this lastExaminedTimeRange variable is the time range with which
+    // any existing time range merges with, if necessary.
+    TimeRange lastExaminedTimeRange = timeRange;
+
+    for (int i = 0; i < allTimeRanges.size(); i++) {
+      TimeRange currentRange = allTimeRanges.get(i);
```
talvarez0507: Doing this loop by referring to an index was O(n) with the arraylist but is now O(n^2) with a linked list. I fixed this with a commit.

PR #38, opened 2020-07-28
Longest task first with task split-up logic

hollyyuqizheng:
I updated the longest task first algorithm so that long tasks that cannot be scheduled into one time range can be broken into smaller segments and be scheduled this way. I also added a new method to the TimeRangeGroup interface.

### DIFF
--- a/portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java
+++ b/portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java
@@ -46,11 +46,7 @@ public class LongestTaskFirstScheduler implements TaskScheduler {
 
     for (Task task : tasksList) {
       List<ScheduledTask> currentScheduledTasks = scheduleOneTask(task, availableTimesGroup);
-
-      currentScheduledTasks.forEach(
-          (scheduledTask) -> {
-            scheduledTasks.add(scheduledTask);
-          });
+      currentScheduledTasks.forEach(scheduledTasks::add);
     }
 
     return scheduledTasks;
@@ -65,10 +61,7 @@ public class LongestTaskFirstScheduler implements TaskScheduler {
   private List<TimeRange> constructAvailableTimeRanges(TimeRangeGroup updatedTimeRangeGroup) {
     Iterator<TimeRange> updatedAvailableTimesGroupIterator = updatedTimeRangeGroup.iterator();
     List<TimeRange> updatedAvailableTimes = new ArrayList();
-    while (updatedAvailableTimesGroupIterator.hasNext()) {
-      updatedAvailableTimes.add(updatedAvailableTimesGroupIterator.next());
-    }
-
+    updatedTimeRangeGroup.iterator().forEachRemaining(updatedAvailableTimes::add);
     return updatedAvailableTimes;
   }
 
@@ -100,7 +93,6 @@ public class LongestTaskFirstScheduler implements TaskScheduler {
 
     for (TimeRange currentFreeTimeRange : currentAvailableTimes) {
       if (taskDuration.getSeconds() == 0) {
-        checkForTaskRenaming(newScheduledTasks);
         return newScheduledTasks;
       }
 
@@ -108,7 +100,6 @@ public class LongestTaskFirstScheduler implements TaskScheduler {
 
       // Reconstruct the task segment's name.
       String taskName = task.getName() + " (Part " + taskSegmentCount + ")";
-      taskSegmentCount++;
 
       String taskDescription = task.getDescription().orElse("");
       TaskPriority taskPriority = task.getPriority();
@@ -116,7 +107,7 @@ public class LongestTaskFirstScheduler implements TaskScheduler {
       // If task's current duration is longer than the free time's,
       // this means the entirety of the free time range is scheduled to this task.
       if (taskDuration.compareTo(currentFreeTimeRange.duration()) > 0) {
-        // Constructs a new taks object with the current free time range's duration.
+        // Constructs a new task object with the current free time range's duration.
         Task taskWithActualScheduledDuration =
             new Task(taskName, taskDescription, currentFreeTimeRange.duration(), taskPriority);
         TimeRange scheduledTaskTimeRange =
@@ -129,6 +120,13 @@ public class LongestTaskFirstScheduler implements TaskScheduler {
         taskDuration = taskDuration.minus(scheduledTaskTimeRange.duration());
       } else {
         // Otherwise, only part of the free time range is needed to schedule this task.
+        // In this case, if the count of segment is 1, then the task can be scheduled in its
+        // entirety.
+        // Retrieve the task's original name without the "(Part 1)" suffix.
+        if (taskSegmentCount == 1) {
+          taskName = task.getName();
+        }
+
         Task taskWithActualScheduledDuration =
             new Task(taskName, taskDescription, taskDuration, taskPriority);
         TimeRange scheduledTaskTimeRange =
@@ -141,9 +139,10 @@ public class LongestTaskFirstScheduler implements TaskScheduler {
 
         // This is also the last free time range that is needed to complete the scheduling
         // for the current task.
-        checkForTaskRenaming(newScheduledTasks);
         return newScheduledTasks;
       }
+
+      taskSegmentCount++;
     }
     // If iterating through all current time ranges finishes, and the task still isn't
     // completely scheduled, then this task cannot be completely scheduled.
@@ -170,38 +169,5 @@ public class LongestTaskFirstScheduler implements TaskScheduler {
     availableTimesGroup.deleteTimeRange(scheduledTaskTimeRange);
 
     return scheduledTaskTimeRange;
-  }
-
-  /**
-   * This method is called when a task can be completely scheduled before returning that scheduled
-   * task. Checks to see if this scheduled task needs to be renamed back to its original name. If
-   * there is only one task in the newly scheduled list of tasks, then this task is completely
-   * scheduled within one time range, so the name should not include "(Part 1)".
-   */
-  private void checkForTaskRenaming(List<ScheduledTask> newScheduledTasks) {
-    if (newScheduledTasks.size() == 1) {
-      ScheduledTask scheduledTask = newScheduledTasks.get(0);
-      ScheduledTask scheduledTaskWithOriginalName = changeTaskNameToOriginal(scheduledTask);
-      newScheduledTasks.set(0, scheduledTaskWithOriginalName);
-    }
-  }
-
-  /** Creates a new ScheduledTask with "(Part 1)" removed. */
-  static ScheduledTask changeTaskNameToOriginal(ScheduledTask scheduledTask) {
-    Task task = scheduledTask.getTask();
-    String currentScheduledTaskName = task.getName();
-    String taskOriginalName = currentScheduledTaskName.split(" \\(Part ")[0];
-
-    Task taskWithOriginalName =
-        new Task(
-            taskOriginalName,
-            task.getDescription().orElse(""),
-            task.getDuration(),
-            task.getPriority());
-
-    ScheduledTask scheduledTaskWithNewName =
-        new ScheduledTask(taskWithOriginalName, scheduledTask.getStartTime());
-
-    return scheduledTaskWithNewName;
   }
 }
--- a/portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java
+++ b/portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java
@@ -6,7 +6,6 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
-import java.util.Iterator;
 import java.util.List;
 
 /** This class models a scheduling algorithm that prioritizes scheduling longer tasks first. */
@@ -59,7 +58,6 @@ public class LongestTaskFirstScheduler implements TaskScheduler {
 
   /** Constructs a list for the updated available time ranges after deletion. */
   private List<TimeRange> constructAvailableTimeRanges(TimeRangeGroup updatedTimeRangeGroup) {
-    Iterator<TimeRange> updatedAvailableTimesGroupIterator = updatedTimeRangeGroup.iterator();
     List<TimeRange> updatedAvailableTimes = new ArrayList();
     updatedTimeRangeGroup.iterator().forEachRemaining(updatedAvailableTimes::add);
     return updatedAvailableTimes;
--- a/portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java
+++ b/portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java
@@ -47,10 +47,11 @@ public class LongestTaskFirstScheduler implements TaskScheduler {
 
     for (Task task : tasksList) {
 
-      Optional<TimeRange> timeRangeForTask = findTimeForTask(task, availableTimesGroup);
+      Optional<TimeRange> timeRangeForTask = findOneTimeRangeForTask(task, availableTimesGroup);
 
       if (!timeRangeForTask.isPresent()) {
         // Try to break up the current task and schedule it as much as possible.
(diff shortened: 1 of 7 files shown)


portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java
```diff
@@ -33,32 +33,37 @@
...
       Optional<TimeRange> timeRangeForTask = findTimeForTask(task, availableTimesGroup);
 
       if (!timeRangeForTask.isPresent()) {
-        continue;
+        // If the total duration of available free times is longer than the task's duration,
+        // then this task can be split up into smaller blocks to be scheduled.
+        // Otherwise, there is not enough free time in the day for this task, so this
+        // task will not be scheduled.
+        if (availableTimesGroup.getTotalDuration().compareTo(task.getDuration()) > 0) {
+          splitUpTaskToSchedule(task, availableTimesGroup, scheduledTasks);
```
jamesnam-google: I would suggest that we use some helper methods to encapsulate the task split-up logic. Also, maybe we can implement a behavior where if there is not enough free time in the day for the task, we dedicate all the free time ranges for this task. The rationale can be that in this case, the user has to spend all the free time on the task, even if it does not get completed.

Consider an approach like,

...
if (!timeRangeForTask.isPresent()) {
  List<TimeRange> timeRangesForTask = findTimeRangesForTask(task, availableTimesGroup);
  // Iterate over timeRangesForTask, and for each time range, construct ScheduledTask, add it,
  // and remove the time range.
} else {
  ...
}

And maybe we can use some suffix for the scheduled task name. For example, the task "Do problem set" can be split into:

Do problem set (session 1/4)
Do problem set (session 2/4)
Do problem set (session 3/4)
Do problem set (session 4/4)

Furthermore, consider including some additional info in the response about tasks that could not be scheduled, either fully or partially. And we can update the UI to convey such info.

And if we take this approach, I don't think you'll need the getTotalDuration method above.

jamesnam-google (changes requested):

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java
hollyyuqizheng: I updated the `splitUpTaskToSchedule` method so that tasks can be partially scheduled. This method now returns a boolean value, but I haven't made the changes corresponding to the UI updates yet, which will come in a future PR. I can imagine there needs to be a lot of client side code change to update the partially scheduled tasks, so I will finalize the logic on the Java end first. The UI updates will come through another PR soon after this one is merged.

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java
```diff
@@ -101,4 +113,69 @@ public SchedulingAlgorithmType getSchedulingAlgorithmType() {
...
+   * @return a boolean value representing whether or not the task is complete scheduled.
+   */
+  private boolean splitUpTaskToSchedule(
+      Task task, TimeRangeGroup availableTimesGroup, List<ScheduledTask> scheduledTasks) {
+    List<TimeRange> currentAvailableTimes = constructAvailableTimeRanges(availableTimesGroup);
+    Duration taskDuration = task.getDuration();
+    int taskSegmentCount = 1;
+
+    for (TimeRange currentFreeTimeRange : currentAvailableTimes) {
+      if (taskDuration.getSeconds() > 0) {
```
jamesnam-google: Can we return early if taskDuration.getSeconds() <= 0 ? Then we can un-indent the lines below.

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java
```diff
@@ -101,4 +113,69 @@ public SchedulingAlgorithmType getSchedulingAlgorithmType() {
...
+        Instant scheduledTime = currentFreeTimeRange.start();
+
+        // Reconstruct the task segment's name.
+        String taskName = task.getName() + " (Part " + taskSegmentCount + ")";
```
jamesnam-google: I think you can do taskSegmentCount++ here

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java
```diff
@@ -101,4 +113,69 @@ public SchedulingAlgorithmType getSchedulingAlgorithmType() {
...
+        String taskDescription = "";
+        if (task.getDescription().isPresent()) {
```
jamesnam-google: We can do

String taskDescription = task.getDescription().orElse("");

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java
```diff
@@ -101,4 +113,69 @@ public SchedulingAlgorithmType getSchedulingAlgorithmType() {
...
+              new Task(taskName, taskDescription, currentFreeTimeRange.duration(), taskPriority);
+          ScheduledTask scheduledTask =
+              new ScheduledTask(taskWithActualScheduledDuration, scheduledTime);
+          scheduledTasks.add(scheduledTask);
+
+          availableTimesGroup.deleteTimeRange(currentFreeTimeRange);
+          taskDuration = taskDuration.minus(currentFreeTimeRange.duration());
+        } else {
+          // Otherwise, only part of the free time range is needed to schedule this task.
+          Task taskWithActualScheduledDuration =
```
jamesnam-google: I think we might be able to use a helper method that can be used in both if and else cases.

ex:
private void scheduleTask(Task task, Instant scheduledTime) {
  ScheduledTask scheduledTask = ...;
  scheduledTasks.add(scheduledTask);
  Instant scheduledTaskEndTime = ...;
  TimeRange scheduledTaskTimeRange = ...;
  availableTimesGroup.deleteTimeRange(...);
  return ScheduledTask;
}

Also, I think we should update ScheduledTask so it exposes a method that returns the scheduled task time range and scheduled task end time. Then we can simplify the helper method further.

Then in the if/else branches we can do,

ScheduledTask scheduledTask = scheduleTask(task, ...);
taskDuration = taskDuration.minus(task.getDuration());

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java
```diff
@@ -33,32 +33,33 @@
...
 
     for (Task task : tasksList) {
-      Duration taskDuration = task.getDuration();
+
       Optional<TimeRange> timeRangeForTask = findTimeForTask(task, availableTimesGroup);
 
       if (!timeRangeForTask.isPresent()) {
-        continue;
+        // Try to break up the current task and schedule it as much as possible.
+        boolean isTaskCompletelyScheduled =
```
jamesnam-google: Consider adding a TODO statement to handle isTaskCompletelyScheduled. I don't think it's used right now.

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java
```diff
@@ -101,4 +113,69 @@ public SchedulingAlgorithmType getSchedulingAlgorithmType() {
 
     return scheduledTimeRangeOptional;
   }
+
+  /**
+   * Keeps splitting up the task to schedule the currently avaible free time ranges until all of the
```
jamesnam-google: Consider updating the method comment to indicate that it updates scheduledTasks.

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java
```diff
@@ -101,4 +113,69 @@ public SchedulingAlgorithmType getSchedulingAlgorithmType() {
...
+   * task is scheduled across the different time ranges. When this method is called, the task may or
+   * may not be complete scheduled.
+   *
+   * @return a boolean value representing whether or not the task is complete scheduled.
```
jamesnam-google: complete scheduled => completely scheduled

jamesnam-google (approved):

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java
```diff
@@ -33,32 +33,34 @@
...
     List<TimeRange> availableTimes = calendarEventsGroup.getFreeTimeRanges();
     TimeRangeGroup availableTimesGroup = new ArrayListTimeRangeGroup(availableTimes);
 
     List<ScheduledTask> scheduledTasks = new ArrayList<ScheduledTask>();
 
     for (Task task : tasksList) {
-      Duration taskDuration = task.getDuration();
-      Optional<TimeRange> timeRangeForTask = findTimeForTask(task, availableTimesGroup);
+
+      Optional<TimeRange> timeRangeForTask = findOneTimeRangeForTask(task, availableTimesGroup);
```
rgcase: This logic seems a little awkward to me. It's surprising to me that we need different logic for handling the cases where the task is scheduled in one range, or when it's split up. Could we just return a `List<TimeRange>` from this method, where it does all the works to decide which `TimeRanges` the task will fill, then the part from lines 59-62 can just happen for each value in the list? It seems more natural to me that way.

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java
```diff
@@ -33,32 +33,34 @@
...
 
       if (!timeRangeForTask.isPresent()) {
-        continue;
+        // Try to break up the current task and schedule it as much as possible.
+        // TODO(hollyyuqizheng): use this boolean value for UI updates.
+        boolean isTaskCompletelyScheduled =
+            splitUpTaskToSchedule(task, availableTimesGroup, scheduledTasks);
+      } else {
+        // This is the case that the current task can be scheduled entirely to one free time range.
+        TimeRange scheduledTimeRange = timeRangeForTask.get();
```
jamesnam-google: Can we call scheduleTaskSegment for this else branch?

jamesnam-google (approved):

raulcruise (approved):

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java
```diff
@@ -33,32 +32,25 @@
...
-
-      // Create and append scheduled tasks.
-      TimeRange scheduledTimeRange = timeRangeForTask.get();
-      ScheduledTask scheduledTask = new ScheduledTask(task, scheduledTimeRange.start());
-      scheduledTasks.add(scheduledTask);
+      List<ScheduledTask> currentScheduledTasks = scheduleOneTask(task, availableTimesGroup);
 
-      // Delete the scheduled time range from the time range group.
-      availableTimesGroup.deleteTimeRange(scheduledTimeRange);
+      currentScheduledTasks.forEach(
```
rgcase: nit: I think you can just do `currentScheduledTasks.forEach(scheduledTasks.add);`

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java
```diff
@@ -69,36 +61,147 @@ public SchedulingAlgorithmType getSchedulingAlgorithmType() {
     return SchedulingAlgorithmType.LONGEST_TASK_FIRST;
   }
 
+  /** Constructs a list for the updated available time ranges after deletion. */
+  private List<TimeRange> constructAvailableTimeRanges(TimeRangeGroup updatedTimeRangeGroup) {
+    Iterator<TimeRange> updatedAvailableTimesGroupIterator = updatedTimeRangeGroup.iterator();
```
rgcase: I think you can make this simpler. You should be able to do:
```
List<TimeRange> updatedAvailableTimes = new ArrayList<>();
updatedTimeRangeGroup.iterator().forEachRemaining(updatedAvailableTimes::add);
return updatedAvailableTimes;
```

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java
```diff
@@ -69,36 +61,147 @@ public SchedulingAlgorithmType getSchedulingAlgorithmType() {
...
-      // Construct and return the scheduled time range based on the start time of this current time
-      // range.
       Instant scheduledTime = currentFreeTimeRange.start();
-      TimeRange scheduledTimeRange =
-          TimeRange.fromStartEnd(scheduledTime, scheduledTime.plus(taskDuration));
-      scheduledTimeRangeOptional = Optional.of(scheduledTimeRange);
-      break;
+
+      // Reconstruct the task segment's name.
+      String taskName = task.getName() + " (Part " + taskSegmentCount + ")";
```
rgcase, line 100: Won't this mean you'll have this "(Part 1)" even if the task fits entirely in one free time range?

EDIT: Oh now I see below how you fix it if the task is scheduled in one time range. I wonder if it's cleaner to instead just move this `taskName` computation inside the `if-else` block. I think the logic can be that if you're inside the `else` block and `taskSegmentCount == ` then you can *not* add the "(Part 1)" suffix, otherwise you do add it?

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java
```diff
@@ -69,36 +61,147 @@ public SchedulingAlgorithmType getSchedulingAlgorithmType() {
...
+      taskSegmentCount++;
+
+      String taskDescription = task.getDescription().orElse("");
+      TaskPriority taskPriority = task.getPriority();
+
+      // If task's current duration is longer than the free time's,
+      // this means the entirety of the free time range is scheduled to this task.
+      if (taskDuration.compareTo(currentFreeTimeRange.duration()) > 0) {
+        // Constructs a new taks object with the current free time range's duration.
```
rgcase: nit: typo taks => task

rgcase (changes requested):

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java
hollyyuqizheng: I see! I moved the taskName computation to the else block.

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java
rgcase: nit: You can remove line 62 now I think?

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java
rgcase: I don't think I see this change. Is there another commit you still have to push?

rgcase (approved):

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java
rgcase: I see the change on line 126 now. That seems good thanks.

PR #41, opened 2020-07-30
LinkedList implementation of TimeRangeGroup

talvarez0507:
This is the linked list implementation of TimeRangeGroup. It uses a linear search for hasTimeRange and edits the list in place for adding and deleting. The tests are exactly the same as the tests for the ArrayList implementation.


portfolio/src/test/java/com/google/sps/LinkedListTimeRangeGroupTest.java
```diff
@@ -0,0 +1,557 @@
+package com.google.sps.data;
```
talvarez0507: These are exactly the same tests as ArrayListTimeRangeGroupTest.java which is already in master except this tests LinkedListTimeRangeGroup instead.

hollyyuqizheng (approved):

raulcruise (approved):

portfolio/src/main/java/com/google/sps/data/LinkedListTimeRangeGroup.java
```diff
@@ -0,0 +1,187 @@
...
+import java.time.Instant;
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.ListIterator;
+
+/**
+ * Models an implementation of the TimeRangeGroup model using LinkedList. This is a slightly
+ * modified version of what hollyyuqizheng wrote for the ArrayList implementation.
```
jamesnam-google: I don't think you need to mention hollyyuqizheng. You can write that it modified ArrayListTime...

portfolio/src/main/java/com/google/sps/data/LinkedListTimeRangeGroup.java
```diff
@@ -0,0 +1,187 @@
...
+ */
+public class LinkedListTimeRangeGroup implements TimeRangeGroup {
+
+  // This list of all the time ranges will be sorted by start time ascending
+  // in add and delete methods. The time ranges stored in this list are
+  // pair-wise disjoint at any moment.
+  public List<TimeRange> allTimeRanges;
```
jamesnam-google: I noticed that this class and ArrayListTimeRangeGroup share a fair amount of code. I would suggest that you use an abstract class to avoid duplicating the code.

ex:
abstract class AbstractListTimeRangeGroup implements TimeRangeGroup {
  protected List<TimeRange> allTimeRanges;

 ...
}

public class LinkedListTimeRangeGroup extends AbstractListTimeRangeGroup implements TimeRangeGroup {
  ...
}

portfolio/src/main/java/com/google/sps/data/LinkedListTimeRangeGroup.java
```diff
@@ -0,0 +1,187 @@
...
+
+    while (iterator.hasNext()) {
+      TimeRange currentRange = iterator.next();
+
+      // If the current range is completely contained by the lastExaminedTimeRange,
+      // we need to remove the current range from the list as lastExaminedTimeRange
+      // will eventually replace it.
+      if (lastExaminedTimeRange.contains(currentRange)) {
+        iterator.remove();
+        continue;
```
jamesnam-google: Consider the case where the initial internal time range list just has 1pm-2pm, and timeRange is 12pm-3pm. Then I think this line gets triggered, and the internal list stays at 1pm-2pm, rather than being updated to 12pm-3pm.

portfolio/src/main/java/com/google/sps/data/LinkedListTimeRangeGroup.java
```diff
@@ -0,0 +1,187 @@
...
+      } else if (currentRange.end().isAfter(lastExaminedTimeRange.end())) {
+        // This is the case that lastExaminedTimeRange and the current time range do not overlap.
+        // If the current range from the original list ends after the last examined time range,
+        // add the time range pointed to by lastExaminedTimeRange to the new list before the current
+        // range,and change the lastExaminedTimeRange to point to the current range.
+        iterator.previous();
+        iterator.add(lastExaminedTimeRange);
+        iterator.next();
+        lastExaminedTimeRange = currentRange;
+      }
```
jamesnam-google, line 79: Should we consider the else case? It can get triggered for example if the internal list was just 1pm-2pm, and timeRange is 3pm-4pm.

portfolio/src/main/java/com/google/sps/data/LinkedListTimeRangeGroup.java
```diff
@@ -0,0 +1,187 @@
...
+        break;
+      }
+    }
+  }
+
+  /**
+   * Helper method for merging two time ranges. This method is package-private so that it can be
+   * tested.
+   */
+  static TimeRange mergeTwoTimeRanges(TimeRange a, TimeRange b) {
```
jamesnam-google: This method can be a helper method defined in the abstract class.

portfolio/src/main/java/com/google/sps/data/LinkedListTimeRangeGroup.java
```diff
@@ -0,0 +1,187 @@
...
+      newTimeRangeEnd = b.end();
+    } else {
+      newTimeRangeEnd = a.end();
+    }
+
+    return TimeRange.fromStartEnd(newTimeRangeStart, newTimeRangeEnd);
+  }
+
+  /** Returns an iterator for the list of all time ranges. */
+  public Iterator<TimeRange> getAllTimeRangesIterator() {
```
jamesnam-google: The abstract class can have this method.

portfolio/src/main/java/com/google/sps/data/LinkedListTimeRangeGroup.java
```diff
@@ -0,0 +1,187 @@
...
+   * <p>Another example for deleting overlapping time ranges: if [3 - 4] and [5 - 6] are in the
+   * original list, deleting [3:30 - 5:30] will result in two new ranges: [3 - 3:30] and [5:30 - 6].
+   */
+  public void deleteTimeRange(TimeRange timeRangeToDelete) {
+
+    ListIterator<TimeRange> iterator = allTimeRanges.listIterator();
+
+    while (iterator.hasNext()) {
+      TimeRange currentRange = iterator.next();
+      if (currentRange.overlaps(timeRangeToDelete)) {
```
jamesnam-google: It might be simpler to continue in the opposite case, so you don't have to indent the inner block.

portfolio/src/main/java/com/google/sps/data/LinkedListTimeRangeGroup.java
```diff
@@ -0,0 +1,187 @@
...
+   * overlap with any other existing time ranges.
+   *
+   * <p>For example, if the list contains [3:00 - 4:00] and [5:00 - 6:00], deleting [3:15 - 3:30]
+   * will result in [3 - 3:15] and [3:30 - 4] as new time ranges.
+   *
+   * <p>Another example for deleting overlapping time ranges: if [3 - 4] and [5 - 6] are in the
+   * original list, deleting [3:30 - 5:30] will result in two new ranges: [3 - 3:30] and [5:30 - 6].
+   */
+  public void deleteTimeRange(TimeRange timeRangeToDelete) {
+
```
jamesnam-google: Unnecessary line break

portfolio/src/main/java/com/google/sps/data/LinkedListTimeRangeGroup.java
```diff
@@ -0,0 +1,187 @@
...
+  /**
+   * Checks if a time range exists in the collection. For example, if [3:00 - 4:00] is in the
+   * collection, [3:00 - 3:30] is considered to exist as a time range in the collection. This method
+   * uses linear search to find the time ranges whose start time is before the target range's start
+   * and whose end time is after the target range's end. Then the method calls contains to see if
+   * the target range is contained within this current range.
+   */
+  public boolean hasTimeRange(TimeRange timeRangeToCheck) {
+
```
jamesnam-google: Unnecessary line break

portfolio/src/test/java/com/google/sps/LinkedListTimeRangeGroupTest.java
```diff
@@ -0,0 +1,557 @@
...
+import java.util.Iterator;
+import java.util.LinkedList;
+import java.util.List;
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public final class LinkedListTimeRangeGroupTest {
```
jamesnam-google: Consider using a parameterized test, which runs the same test suite on LinkedListTimeRangeGroup as well as ArrayListTimeRangeGroup.

Reference:
https://junit.org/junit4/javadoc/4.12/org/junit/runners/Parameterized.html

Maybe we can name it as TimeRangeGroupTest

jamesnam-google (changes requested):

portfolio/src/main/java/com/google/sps/data/LinkedListTimeRangeGroup.java
```diff
@@ -0,0 +1,187 @@
...
+  // pair-wise disjoint at any moment.
+  public List<TimeRange> allTimeRanges;
+
+  /**
+   * Adds all the input time ranges to the list of all time ranges. Also sorts the list of all
+   * ranges in the constructor.
+   */
+  public LinkedListTimeRangeGroup(Collection<TimeRange> timeRanges) {
+    allTimeRanges = new LinkedList<TimeRange>();
+    timeRanges.forEach(
```
rgcase: nit: You can just do `timeRanges.forEach(addTimeRange)`.

portfolio/src/main/java/com/google/sps/data/LinkedListTimeRangeGroup.java
rgcase: nit: It's named `.iterator()` in the other implementation and you should update yours to match, so it can be used in a foreach loop.

rgcase (changes requested):

portfolio/src/main/java/com/google/sps/data/LinkedListTimeRangeGroup.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/data/LinkedListTimeRangeGroup.java
talvarez0507: Done. I will change the ArrayList implementation to extend AbstractListTimeRangeGroup in another PR.

portfolio/src/main/java/com/google/sps/data/LinkedListTimeRangeGroup.java
talvarez0507: Fixed. I had to rewrite a bit of the code but all of these corner cases having to due with a single element or adding one element all the way at the end work fine now. Tests were also added for these cases.

portfolio/src/main/java/com/google/sps/data/LinkedListTimeRangeGroup.java
talvarez0507: The else case means that the currentRange and lastExaminedTimeRange do not overlap and that lastExaminedTimeRange is strictly later than currentRange. On some further iteration the else if will trigger and insert it into the list once it comes before something in the list so we don't have to worry about the else case.

portfolio/src/main/java/com/google/sps/data/LinkedListTimeRangeGroup.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/data/LinkedListTimeRangeGroup.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/data/LinkedListTimeRangeGroup.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/data/LinkedListTimeRangeGroup.java
talvarez0507: Removed.

portfolio/src/main/java/com/google/sps/data/LinkedListTimeRangeGroup.java
talvarez0507: Removed.

portfolio/src/test/java/com/google/sps/LinkedListTimeRangeGroupTest.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/data/LinkedListTimeRangeGroup.java
talvarez0507: I was not able to make this work. I got the error message "cannot find symbol addTimeRange" even though it is a method in the class.

portfolio/src/main/java/com/google/sps/data/AbstractListTimeRangeGroup.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.List;
+
+abstract class AbstractListTimeRangeGroup implements TimeRangeGroup {
+  protected List<TimeRange> allTimeRanges;
+
+  /**
+   * Helper method for merging two time ranges. This method is package-private so that it can be
+   * tested.
+   */
+  public static TimeRange mergeTwoTimeRanges(TimeRange a, TimeRange b) {
```
jamesnam-google: I think this method can be protected instead of public.

portfolio/src/main/java/com/google/sps/data/AbstractListTimeRangeGroup.java
```diff
@@ -0,0 +1,43 @@
...
+      newTimeRangeEnd = b.end();
+    } else {
+      newTimeRangeEnd = a.end();
+    }
+
+    return TimeRange.fromStartEnd(newTimeRangeStart, newTimeRangeEnd);
+  }
+
+  /** Returns an iterator for the list of all time ranges. */
+  public Iterator<TimeRange> iterator() {
```
jamesnam-google, line 41: (Optional) You can add the @Override annotation here (capital O, which somehow is not capitalized when I submit the comment on GitHub).

Reference:
https://docs.oracle.com/javase/tutorial/java/annotations/predefined.html

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
```diff
@@ -0,0 +1,146 @@
...
+        (range) -> {
+          addTimeRange(range);
+        });
+  }
+
+  /**
+   * Adds a new time range to the list. If the time range to add overlaps with any existing time
+   * range, the overlapping time ranges will be merged.
+   */
+  public void addTimeRange(TimeRange timeRange) {
```
jamesnam-google, line 24: You can consider adding the override annotation

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
```diff
@@ -0,0 +1,146 @@
...
+/** Models an implementation of the TimeRangeGroup model using ArrayList. */
+public class ArrayListTimeRangeGroup extends AbstractListTimeRangeGroup implements TimeRangeGroup {
+
+  /**
+   * Adds all the input time ranges to the list of all time ranges. Also sorts the list of all
+   * ranges in the constructor.
+   */
+  public ArrayListTimeRangeGroup(Collection<TimeRange> timeRanges) {
+    allTimeRanges = new ArrayList<TimeRange>();
+    timeRanges.forEach(
```
jamesnam-google: Consider pulling out this forEach into a helper method. I think it can be a protected method declared in the abstract class. Then here you can invoke it.

(As a result, you may have to declare addTimeRange in the abstract method.)

portfolio/src/test/java/com/google/sps/TimeRangeGroupTest.java
```diff
@@ -0,0 +1,669 @@
...
+import org.junit.runners.Parameterized.Parameters;
+
+@RunWith(value = Parameterized.class)
+public class TimeRangeGroupTest {
+  @Parameters
+  public static List parameters() {
+    return Arrays.asList(ArrayListTimeRangeGroup.class, LinkedListTimeRangeGroup.class);
+  }
+
+  private Class timeRangeGroupClass;
```
jamesnam-google: I think you can do,

@Parameter Class timeRangeGroupClass;

Then you won't need the TimeRangeGroupTest constructor below.

portfolio/src/main/java/com/google/sps/data/LinkedListTimeRangeGroup.java
```diff
@@ -0,0 +1,157 @@
...
+ */
+public class LinkedListTimeRangeGroup extends AbstractListTimeRangeGroup implements TimeRangeGroup {
+
+  /**
+   * Adds all the input time ranges to the list of all time ranges. Also sorts the list of all
+   * ranges in the constructor.
+   */
+  public LinkedListTimeRangeGroup(Collection<TimeRange> timeRanges) {
+    allTimeRanges = new LinkedList<TimeRange>();
+    timeRanges.forEach(
```
jamesnam-google: As I suggested earlier, I think you can just invoke the helper method here for forEach.

portfolio/src/main/java/com/google/sps/data/LinkedListTimeRangeGroup.java
```diff
@@ -0,0 +1,157 @@
...
+    // linked list can be confusing. Each case should be handled differently
+    // so using iterator.hasNext() and currentRange=iterator.next() in each
+    // iteration of this loop would make things harder to keep track of.
+    while (true) {
+
+      // If the current range is completely contained by the lastExaminedTimeRange,
+      // we need to remove the current range from the list as lastExaminedTimeRange
+      // will eventually replace it.
+      if (lastExaminedTimeRange.contains(currentRange)
+          && !lastExaminedTimeRange.equals(currentRange)) {
```
jamesnam-google, line 57: Do we need the ! equals check?

portfolio/src/main/java/com/google/sps/data/LinkedListTimeRangeGroup.java
```diff
@@ -0,0 +1,157 @@
...
+        iterator.remove();
+        if (iterator.hasNext()) {
+          currentRange = iterator.next();
+        }
+      }
+
+      // The case for when two time ranges need to be merged.
+      // Similar to the case above, it will eventually be replaced so we can
+      // remove currentRange
+      if (lastExaminedTimeRange.overlaps(currentRange)
```
jamesnam-google, line 63: Can this be an else if? If so, then we won't need the ! contains check.

jamesnam-google (approved):

portfolio/src/test/java/com/google/sps/TimeRangeGroupTest.java
```diff
@@ -0,0 +1,669 @@
...
+
+  /** Tests for adding a time range that entirely contains an existing one. */
+  @Test
+  public void testAddSupersetRange() {
+    // Time Ranges:      |-A-|
+    // To add:       |-----B-----|
+    // Result:       |-----------|
+    Instant timeRangeOneStart = Instant.now();
+    Instant timeRangeOneEnd = timeRangeOneStart.plusSeconds(3000);
+    Instant timeRangeTwoStart = timeRangeOneStart.plusSeconds(500);
```
hollyyuqizheng: timeRangeTwoStart should probably be `timeRangeOneStart.minusSeconds(500)` ?

portfolio/src/test/java/com/google/sps/TimeRangeGroupTest.java
```diff
@@ -0,0 +1,669 @@
...
+
+    TimeRange timeRangeOne = TimeRange.fromStartEnd(timeRangeOneStart, timeRangeOneEnd);
+
+    List<TimeRange> originalTimeRanges = Arrays.asList(timeRangeOne);
+    for (TimeRange timeRange : originalTimeRanges) {
+      timeRangeGroup.addTimeRange(timeRange);
+    }
+    List<TimeRange> expected = Arrays.asList(timeRangeOne);
+
+    timeRangeGroup.addTimeRange(timeRangeOne);
```
hollyyuqizheng, line 290: Hmmm what if you try to construct a new TimeRange instance called `timeRangeTwo`, and initialize this new time range with the exact start and end times of timeRangeOne? I think in this case the test passes for array list time range group, at least it did when I tried it on my own.

portfolio/src/main/java/com/google/sps/data/AbstractListTimeRangeGroup.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/data/AbstractListTimeRangeGroup.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
talvarez0507: Done.

portfolio/src/test/java/com/google/sps/TimeRangeGroupTest.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/data/LinkedListTimeRangeGroup.java
talvarez0507: Done.

portfolio/src/main/java/com/google/sps/data/LinkedListTimeRangeGroup.java
talvarez0507: If we remove the !equals then when we try to add an event that already exists it is removed and never added.

portfolio/src/main/java/com/google/sps/data/LinkedListTimeRangeGroup.java
talvarez0507: I think this can't be an else if because the if statement that comes before this and the else if statement after this are not mutually exclusive.

portfolio/src/test/java/com/google/sps/TimeRangeGroupTest.java
talvarez0507: Fixed.

portfolio/src/test/java/com/google/sps/TimeRangeGroupTest.java
talvarez0507: Seems like I just had outdated code, ArrayListTimeRangeGroup passes this test just fine now.

rgcase (approved):

PR #43, opened 2020-08-03
Highest Priority First Algorithm Implementation

raulcruise:
The HighestPriorityFirst scheduling algorithm prioritizes Tasks with the highest priority, and within each set of Tasks with equal  Priority those with shorter duration are prioritized so that when we reach the end of the availableTimeRanges we can disregard any remaining Tasks of equal priority knowing that they won't fit in any of the previously viewed TimeRanges. A new comparator was also added to the TaskQueue class so that it could sort the tasks in the order which the algorithm requires it.

### DIFF
--- a/portfolio/src/main/java/com/google/sps/data/HighestPriorityFirstScheduler.java
+++ b/portfolio/src/main/java/com/google/sps/data/HighestPriorityFirstScheduler.java
@@ -1,0 +1,95 @@
+package com.google.sps.data;
+
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Optional;
+
+public class HighestPriorityFirstScheduler implements TaskScheduler {
+  private static final Comparator<ScheduledTask> sortByScheduledStartTimeAscending =
+      Comparator.comparing(ScheduledTask::getStartTime);
+
+  /** This method schedules tasks from highest to lowest priority and */
+  public Collection<ScheduledTask> schedule(
+      Collection<CalendarEvent> events,
+      Collection<Task> tasks,
+      Instant workHoursStartTime,
+      Instant workHoursEndTime) {
+
+    List<CalendarEvent> eventsList = new ArrayList<CalendarEvent>(events);
+    List<Task> tasksList = new ArrayList<Task>(tasks);
+
+    TaskQueue taskQueue = new TaskQueue(tasksList, getSchedulingAlgorithmType());
+
+    CalendarEventsGroup calendarEventsGroup =
+        new CalendarEventsGroup(eventsList, workHoursStartTime, workHoursEndTime);
+
+    List<TimeRange> availableTimes = calendarEventsGroup.getFreeTimeRanges();
+    TimeRangeGroup availableTimesGroup = new ArrayListTimeRangeGroup(availableTimes);
+
+    List<ScheduledTask> scheduledTasks = new ArrayList<ScheduledTask>();
+
+    Iterator<TimeRange> availableTimesIterator = availableTimesGroup.iterator();
+    int rangeIndex = 0;
+
+    Instant currentScheduleTime = workHoursStartTime;
+
+    Boolean getNextTimeRange = true;
+    TimeRange availableTimeRange = availableTimes.get(0);
+
+    while (availableTimesIterator.hasNext() && !taskQueue.isEmpty()) {
+      if (getNextTimeRange) {
+        availableTimeRange = availableTimesIterator.next();
+      }
+      Task task = taskQueue.peek();
+
+      if (availableTimeRange.start().isAfter(currentScheduleTime)) {
+        currentScheduleTime = availableTimeRange.start();
+      }
+
+      if (!currentScheduleTime
+          .plusSeconds(task.getDuration().getSeconds())
+          .isAfter(availableTimeRange.end())) {
+        ScheduledTask scheduledTask =
+            new ScheduledTask(task, currentScheduleTime, Optional.of(true));
+        scheduledTasks.add(scheduledTask);
+        availableTimesGroup.deleteTimeRange(
+            TimeRange.fromStartEnd(
+                scheduledTask.getStartTime(),
+                scheduledTask
+                    .getStartTime()
+                    .plusSeconds(scheduledTask.getTask().getDuration().getSeconds())));
+        currentScheduleTime = currentScheduleTime.plusSeconds(task.getDuration().getSeconds());
+        taskQueue.remove();
+        if (!taskQueue.isEmpty()
+            && taskQueue.peek().getPriority().getPriority() != task.getPriority().getPriority()) {
+          availableTimesIterator = availableTimesGroup.iterator();
+          currentScheduleTime = workHoursStartTime;
+          getNextTimeRange = true;
+        } else {
+          getNextTimeRange = false;
+        }
+      } else if (!availableTimesIterator.hasNext()) {
+        taskQueue.remove();
+        if (!taskQueue.isEmpty()
+            && taskQueue.peek().getPriority().getPriority() != task.getPriority().getPriority()) {
+          availableTimesIterator = availableTimesGroup.iterator();
+        }
+      } else {
+        getNextTimeRange = true;
+      }
+    }
+
+    Collections.sort(scheduledTasks, sortByScheduledStartTimeAscending);
+    return scheduledTasks;
+  }
+
+  /** Returns the scheduler's type, which is Highest Priority First. */
+  public SchedulingAlgorithmType getSchedulingAlgorithmType() {
+    return SchedulingAlgorithmType.HIGHEST_PRIORITY_FIRST;
+  }
+}
--- a/portfolio/src/main/java/com/google/sps/data/HighestPriorityFirstScheduler.java
+++ b/portfolio/src/main/java/com/google/sps/data/HighestPriorityFirstScheduler.java
@@ -39,17 +39,17 @@ public class HighestPriorityFirstScheduler implements TaskScheduler {
     Instant currentScheduleTime = workHoursStartTime;
 
     while (currentAvailableTimesIndex < currentAvailableTimes.size() && !taskQueue.isEmpty()) {
-      TimeRange availableTimeRange = currentAvailableTimes.get(currentAvailableTimesIndex);
+      TimeRange currentAvailableTimeRange = currentAvailableTimes.get(currentAvailableTimesIndex);
       Task task = taskQueue.peek();
 
-      if (availableTimeRange.start().isAfter(currentScheduleTime)) {
-        currentScheduleTime = availableTimeRange.start();
+      if (currentAvailableTimeRange.start().isAfter(currentScheduleTime)) {
+        currentScheduleTime = currentAvailableTimeRange.start();
       }
 
       boolean doesTaskFit =
           !currentScheduleTime
               .plusSeconds(task.getDuration().getSeconds())
-              .isAfter(availableTimeRange.end());
+              .isAfter(currentAvailableTimeRange.end());
 
       if (doesTaskFit) {
         ScheduledTask scheduledTask =
--- a/portfolio/src/main/java/com/google/sps/data/HighestPriorityFirstScheduler.java
+++ b/portfolio/src/main/java/com/google/sps/data/HighestPriorityFirstScheduler.java
@@ -28,6 +28,10 @@ public class HighestPriorityFirstScheduler implements TaskScheduler {
     CalendarEventsGroup calendarEventsGroup =
         new CalendarEventsGroup(eventsList, workHoursStartTime, workHoursEndTime);
 
+    // When accessing available TimeRanges, work with currentAvailableTimes
+    // since availableTimes will not be updated after creation while
+    // currentAvailableTimes will be updated everytime the algorithm works
+    // with a new priority value. Updates included modified or deleted TimeRanges.
     List<TimeRange> availableTimes = calendarEventsGroup.getFreeTimeRanges();
     TimeRangeGroup availableTimesGroup = new ArrayListTimeRangeGroup(availableTimes);
     List<TimeRange> currentAvailableTimes = getAvailableTimeRangesList(availableTimesGroup);
--- a/portfolio/src/main/java/com/google/sps/data/HighestPriorityFirstScheduler.java
+++ b/portfolio/src/main/java/com/google/sps/data/HighestPriorityFirstScheduler.java
@@ -5,7 +5,6 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Optional;
 
@@ -30,25 +29,16 @@ public class HighestPriorityFirstScheduler implements TaskScheduler {
 
     List<TimeRange> availableTimes = calendarEventsGroup.getFreeTimeRanges();
     TimeRangeGroup availableTimesGroup = new ArrayListTimeRangeGroup(availableTimes);
+    List<TimeRange> currentAvailableTimes = availableTimeRangesList(availableTimesGroup);
 
     List<ScheduledTask> scheduledTasks = new ArrayList<ScheduledTask>();
 
-    Iterator<TimeRange> availableTimesIterator = availableTimesGroup.iterator();
+    int availableTimesIndex = 0;
 
     Instant currentScheduleTime = workHoursStartTime;
 
-    Boolean getNextTimeRange = true;
-
-    // Set the first availableTimeRange using the first list value
-    // in order to avoid getting the value from the availableTimesIterator
-    // which would prevent the while loop from running if there were to
-    // only be a single availableTimeRange.
-    TimeRange availableTimeRange = availableTimes.get(0);
-
-    while (availableTimesIterator.hasNext() && !taskQueue.isEmpty()) {
-      if (getNextTimeRange) {
-        availableTimeRange = availableTimesIterator.next();
-      }
+    while (availableTimesIndex < currentAvailableTimes.size() && !taskQueue.isEmpty()) {
+      TimeRange availableTimeRange = currentAvailableTimes.get(availableTimesIndex);
       Task task = taskQueue.peek();
 
       if (availableTimeRange.start().isAfter(currentScheduleTime)) {
@@ -70,32 +60,31 @@ public class HighestPriorityFirstScheduler implements TaskScheduler {
                 scheduledTask
                     .getStartTime()
                     .plusSeconds(scheduledTask.getTask().getDuration().getSeconds())));
+        // After deleting a TimeRange, we must renew the list.
+        currentAvailableTimes = availableTimeRangesList(availableTimesGroup);
 
         // Push the currentScheduleTime back by the amount that was scheduled.
         currentScheduleTime = currentScheduleTime.plusSeconds(task.getDuration().getSeconds());
         taskQueue.remove();
 
         // If the next task's priority is different from the task that was just scheduled,
-        // then reset the availableTimesIterator, currentScheduleTime, and make the loop
+        // then reset the availableTimesIndex, currentScheduleTime, and make the loop
         // grab a TimeRange. We go back to the first availableTimeRange in order to schedule
         // more tasks towards the beginning of the availableTimes.
         if (!taskQueue.isEmpty()
             && taskQueue.peek().getPriority().getPriority() != task.getPriority().getPriority()) {
-          availableTimesIterator = availableTimesGroup.iterator();
+          availableTimesIndex = 0;
           currentScheduleTime = workHoursStartTime;
-          getNextTimeRange = true;
-        } else {
-          // If the task's priority is not different, then we can simply continue running the
-          // the loop without retrieving the next availableTimeRange since there might be
-          // some time left over towards the end of the current one.
-          getNextTimeRange = false;
         }
-      } else if (!availableTimesIterator.hasNext()) {
+        // If the task's priority is not different, then we can simply continue running the
+        // the loop without retrieving the next availableTimeRange since there might be
+        // some time left over towards the end of the current one.
+      } else if (availableTimesIndex == currentAvailableTimes.size() - 1) {
         // If we've reached the end of the availableTimeRange group and we can't schedule a task,
         // then we can remove all the remaining tasks of equal priority since they will all be
         // longer in duration therefore, they will not be able to be scheduled either.
         if (!taskQueue.isEmpty()) {
-          availableTimesIterator = availableTimesGroup.iterator();
+          availableTimesIndex = 0;
           while (!taskQueue.isEmpty()
               && taskQueue.peek().getPriority().getPriority() == task.getPriority().getPriority()) {
             taskQueue.remove();
@@ -104,12 +93,18 @@ public class HighestPriorityFirstScheduler implements TaskScheduler {
       } else {
         // Get the next TimeRange if the Task cannot be scheduled but we aren't at the final
         // TimeRange.
-        getNextTimeRange = true;
+        availableTimesIndex++;
       }
     }
 
     Collections.sort(scheduledTasks, sortByScheduledStartTimeAscending);
     return scheduledTasks;
+  }
+
+  private List<TimeRange> availableTimeRangesList(TimeRangeGroup updatedTimeRangeGroup) {
+    List<TimeRange> availableTimes = new ArrayList();
+    updatedTimeRangeGroup.iterator().forEachRemaining(availableTimes::add);
+    return availableTimes;
   }
 
   /** Returns the scheduler's type, which is Highest Priority First. */
--- a/portfolio/src/main/java/com/google/sps/data/HighestPriorityFirstScheduler.java
+++ b/portfolio/src/main/java/com/google/sps/data/HighestPriorityFirstScheduler.java
@@ -12,7 +12,9 @@ public class HighestPriorityFirstScheduler implements TaskScheduler {
   private static final Comparator<ScheduledTask> sortByScheduledStartTimeAscending =
       Comparator.comparing(ScheduledTask::getStartTime);
 
-  /** This method schedules tasks from highest to lowest priority and */
+  /**
+   * This method schedules tasks from highest to lowest priority and shortest to longest duration.
+   */
   public Collection<ScheduledTask> schedule(
       Collection<CalendarEvent> events,
       Collection<Task> tasks,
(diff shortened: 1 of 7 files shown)


portfolio/src/test/java/com/google/sps/HighestPriorityFirstSchedulerTest.java
```diff
@@ -0,0 +1,309 @@
...
+            task3, SchedulerTestUtil.TIME_0900, SchedulerTestUtil.completelyScheduled);
+
+    Collection<ScheduledTask> actual =
+        highestPriorityFirstScheduler.schedule(
+            events, tasks, SchedulerTestUtil.TIME_0900, SchedulerTestUtil.TIME_1000);
+    Collection<ScheduledTask> expected = Arrays.asList(scheduledTask3, scheduledTask2);
+
+    Assert.assertEquals(expected, actual);
+  }
+}
```
hollyyuqizheng, line 416: I would add a few more tests with tasks of same priority but different duration.

raulcruise: Done.

portfolio/src/main/java/com/google/sps/data/HighestPriorityFirstScheduler.java
```diff
@@ -0,0 +1,114 @@
...
+
+        // Push the currentScheduleTime back by the amount that was scheduled.
+        currentScheduleTime = currentScheduleTime.plusSeconds(task.getDuration().getSeconds());
+        taskQueue.remove();
+
+        // If the next task's priority is different from the task that was just scheduled,
+        // then reset the availableTimesIndex, currentScheduleTime, and make the loop
+        // grab a TimeRange. We go back to the first availableTimeRange in order to schedule
+        // more tasks towards the beginning of the availableTimes.
+        if (!taskQueue.isEmpty()
```
hollyyuqizheng: Maybe it's cleaner to pull this condition checking into a helper method called `isNextTaskSamePriority` ?

portfolio/src/main/java/com/google/sps/data/HighestPriorityFirstScheduler.java
```diff
@@ -0,0 +1,114 @@
...
+        // TimeRange.
+        availableTimesIndex++;
+      }
+    }
+
+    Collections.sort(scheduledTasks, sortByScheduledStartTimeAscending);
+    return scheduledTasks;
+  }
+
+  private List<TimeRange> availableTimeRangesList(TimeRangeGroup updatedTimeRangeGroup) {
```
hollyyuqizheng: I would rename the method into `getAvailableTimeRangesList`.

hollyyuqizheng (approved):
Some comments on method naming.

portfolio/src/main/java/com/google/sps/data/HighestPriorityFirstScheduler.java
raulcruise: Done.

portfolio/src/main/java/com/google/sps/data/HighestPriorityFirstScheduler.java
raulcruise: Done.

talvarez0507 (approved):

portfolio/src/main/java/com/google/sps/data/HighestPriorityFirstScheduler.java
```diff
@@ -0,0 +1,118 @@
...
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Optional;
+
+public class HighestPriorityFirstScheduler implements TaskScheduler {
+  private static final Comparator<ScheduledTask> sortByScheduledStartTimeAscending =
+      Comparator.comparing(ScheduledTask::getStartTime);
+
+  /** This method schedules tasks from highest to lowest priority and */
```
jamesnam-google: This comment seems incomplete.

portfolio/src/main/java/com/google/sps/data/HighestPriorityFirstScheduler.java
```diff
@@ -0,0 +1,118 @@
...
+    List<Task> tasksList = new ArrayList<Task>(tasks);
+
+    TaskQueue taskQueue = new TaskQueue(tasksList, getSchedulingAlgorithmType());
+
+    CalendarEventsGroup calendarEventsGroup =
+        new CalendarEventsGroup(eventsList, workHoursStartTime, workHoursEndTime);
+
+    List<TimeRange> availableTimes = calendarEventsGroup.getFreeTimeRanges();
+    TimeRangeGroup availableTimesGroup = new ArrayListTimeRangeGroup(availableTimes);
+    List<TimeRange> currentAvailableTimes = getAvailableTimeRangesList(availableTimesGroup);
```
jamesnam-google: I'm confused. It seems like you are converting List<TimeRange> to TimeRangeGroup, then back to List<TimeRange>. What's the reasoning here?

portfolio/src/main/java/com/google/sps/data/HighestPriorityFirstScheduler.java
```diff
@@ -0,0 +1,118 @@
...
+
+    List<ScheduledTask> scheduledTasks = new ArrayList<ScheduledTask>();
+
+    int availableTimesIndex = 0;
+
+    Instant currentScheduleTime = workHoursStartTime;
+
+    while (availableTimesIndex < currentAvailableTimes.size() && !taskQueue.isEmpty()) {
```
jamesnam-google: Consider adding some comment on how the algorithm works at a high level.

portfolio/src/main/java/com/google/sps/data/HighestPriorityFirstScheduler.java
```diff
@@ -0,0 +1,118 @@
...
+      TimeRange availableTimeRange = currentAvailableTimes.get(availableTimesIndex);
+      Task task = taskQueue.peek();
+
+      if (availableTimeRange.start().isAfter(currentScheduleTime)) {
+        currentScheduleTime = availableTimeRange.start();
+      }
+
+      if (!currentScheduleTime
```
jamesnam-google: Consider assigning this value to a variable.

ex:
boolean doesTaskFit = !currentScheduleTime...
if (doesTaskFit) {
  ...

portfolio/src/main/java/com/google/sps/data/HighestPriorityFirstScheduler.java
```diff
@@ -0,0 +1,118 @@
...
+          .plusSeconds(task.getDuration().getSeconds())
+          .isAfter(availableTimeRange.end())) {
+        ScheduledTask scheduledTask =
+            new ScheduledTask(task, currentScheduleTime, Optional.of(true));
```
jamesnam-google: Can we have param comment for the third param?

portfolio/src/main/java/com/google/sps/data/HighestPriorityFirstScheduler.java
```diff
@@ -0,0 +1,118 @@
...
+        // Delete the TimeRange that has been scheduled over so that different priority tasks
+        // won't be scheduled over the same time.
+        availableTimesGroup.deleteTimeRange(
+            TimeRange.fromStartEnd(
+                scheduledTask.getStartTime(),
+                scheduledTask
+                    .getStartTime()
+                    .plusSeconds(scheduledTask.getTask().getDuration().getSeconds())));
+        // After deleting a TimeRange, we must renew the list.
+        currentAvailableTimes = getAvailableTimeRangesList(availableTimesGroup);
```
jamesnam-google: This should be unnecessary if we just work with TimeRangeGroup, rather than available time ranges list.

portfolio/src/main/java/com/google/sps/data/HighestPriorityFirstScheduler.java
```diff
@@ -0,0 +1,118 @@
...
+
+        // Push the currentScheduleTime back by the amount that was scheduled.
+        currentScheduleTime = currentScheduleTime.plusSeconds(task.getDuration().getSeconds());
+        taskQueue.remove();
+
+        // If the next task's priority is different from the task that was just scheduled,
+        // then reset the availableTimesIndex, currentScheduleTime, and make the loop
+        // grab a TimeRange. We go back to the first availableTimeRange in order to schedule
+        // more tasks towards the beginning of the availableTimes.
+        if (isNextTaskDifferentPriority(taskQueue, task)) {
```
jamesnam-google, line 88: I'm not sure why this is necessary. But I understand that you want to make use of earlier, available time ranges. I think the following approach might be simpler.

1. First, you order the tasks by their priority (in descending order).
2. For each task, you identify the first time range that can fit that task. And schedule it there.

I think the current implementation is somewhat overly complex. The suggested approach above should be simpler, I think.

jamesnam-google (changes requested):

portfolio/src/main/java/com/google/sps/data/HighestPriorityFirstScheduler.java
raulcruise: Done.

portfolio/src/main/java/com/google/sps/data/HighestPriorityFirstScheduler.java
raulcruise: Done.

portfolio/src/main/java/com/google/sps/data/HighestPriorityFirstScheduler.java
raulcruise: Done.

portfolio/src/main/java/com/google/sps/data/HighestPriorityFirstScheduler.java
raulcruise: The reasoning behind this is that since the tasks are also ordered from shortest to longest, when we're sorting tasks with the same priority we can sort all the possible tasks for a priority in one run through of the availableTimeRanges instead of possibly (worst case scenario)having to iterate to the last availableTimeRange to schedule each task.

portfolio/src/main/java/com/google/sps/data/HighestPriorityFirstScheduler.java
```diff
@@ -0,0 +1,124 @@
...
+        // If the task's priority is not different, then we can simply continue running the
+        // the loop without retrieving the next availableTimeRange since there might be
+        // some time left over towards the end of the current one.
+      } else if (availableTimesIndex == currentAvailableTimes.size() - 1) {
+        // If we've reached the end of the availableTimeRange group and we can't schedule a task,
+        // then we can remove all the remaining tasks of equal priority since they will all be
+        // longer in duration therefore, they will not be able to be scheduled either.
+        if (!taskQueue.isEmpty()) {
+          availableTimesIndex = 0;
+          while (!taskQueue.isEmpty()
```
jamesnam-google: Consider using a helper method for this while loop.

ex:
private void removeTasksWithPriority(TaskQueue taskQueue, TaskPriority taskPriority) {
  ...
}

portfolio/src/main/java/com/google/sps/data/HighestPriorityFirstScheduler.java
```diff
@@ -0,0 +1,124 @@
...
+
+  /**
+   * This method schedules tasks from highest to lowest priority and shortest to longest duration.
+   */
+  public Collection<ScheduledTask> schedule(
+      Collection<CalendarEvent> events,
+      Collection<Task> tasks,
+      Instant workHoursStartTime,
+      Instant workHoursEndTime) {
+
```
jamesnam-google: Unnecessary line break

portfolio/src/main/java/com/google/sps/data/HighestPriorityFirstScheduler.java
```diff
@@ -0,0 +1,124 @@
...
+    List<CalendarEvent> eventsList = new ArrayList<CalendarEvent>(events);
+    List<Task> tasksList = new ArrayList<Task>(tasks);
+
+    TaskQueue taskQueue = new TaskQueue(tasksList, getSchedulingAlgorithmType());
+
+    CalendarEventsGroup calendarEventsGroup =
+        new CalendarEventsGroup(eventsList, workHoursStartTime, workHoursEndTime);
+
+    List<TimeRange> availableTimes = calendarEventsGroup.getFreeTimeRanges();
```
jamesnam-google, line 31: Maybe add a comment that says that we should work with currentAvailableTimes and not use availableTimes.

portfolio/src/main/java/com/google/sps/data/HighestPriorityFirstScheduler.java
```diff
@@ -0,0 +1,124 @@
...
+    TimeRangeGroup availableTimesGroup = new ArrayListTimeRangeGroup(availableTimes);
+    List<TimeRange> currentAvailableTimes = getAvailableTimeRangesList(availableTimesGroup);
+
+    List<ScheduledTask> scheduledTasks = new ArrayList<ScheduledTask>();
+
+    int availableTimesIndex = 0;
```
jamesnam-google, line 37: Consider renaming to currentAvailableTimesIndex, given that you're using this variable to access currentAvailableTimes and not availableTimes

portfolio/src/main/java/com/google/sps/data/HighestPriorityFirstScheduler.java
```diff
@@ -0,0 +1,124 @@
...
+
+    Instant currentScheduleTime = workHoursStartTime;
+
+    while (availableTimesIndex < currentAvailableTimes.size() && !taskQueue.isEmpty()) {
+      TimeRange availableTimeRange = currentAvailableTimes.get(availableTimesIndex);
```
jamesnam-google: Consider renaming to currentAvailableTimeRange

portfolio/src/main/java/com/google/sps/data/HighestPriorityFirstScheduler.java
```diff
@@ -0,0 +1,124 @@
...
+      Task task = taskQueue.peek();
+
+      if (availableTimeRange.start().isAfter(currentScheduleTime)) {
+        currentScheduleTime = availableTimeRange.start();
+      }
+
+      Boolean doesTaskFit =
```
jamesnam-google: We can use boolean rather than Boolean

jamesnam-google (changes requested):

portfolio/src/main/java/com/google/sps/data/HighestPriorityFirstScheduler.java
raulcruise: Done.

portfolio/src/main/java/com/google/sps/data/HighestPriorityFirstScheduler.java
raulcruise: Done.

portfolio/src/main/java/com/google/sps/data/HighestPriorityFirstScheduler.java
raulcruise: Done.

portfolio/src/test/java/com/google/sps/TaskQueueTest.java
```diff
@@ -73,4 +72,146 @@ public void checkCorrectSortingShortestTaskFirst() {
...
+        new Task("Task 5", "Fifth task", DURATION_30_MINUTES, SchedulerTestUtil.PRIORITY_THREE);
+
+    List<Task> taskList = Arrays.asList(task3, task1, task2, task5, task4);
+
+    List<Task> expected = Arrays.asList(task2, task4, task5, task3, task1);
+    List<Task> actual = new ArrayList<Task>();
+
+    TaskQueue taskQueue = new TaskQueue(taskList, HIGHEST_PRIORITY_FIRST);
+
+    for (int i = 0; i < expected.size(); i++) {
```
jamesnam-google: It might be safer to use a while loop to go through taskQueue completely.

portfolio/src/test/java/com/google/sps/TaskQueueTest.java
```diff
@@ -73,4 +72,146 @@ public void checkCorrectSortingShortestTaskFirst() {
...
+            SchedulerTestUtil.PRIORITY_ONE);
+
+    List<Task> taskList = Arrays.asList(task3, task1, task2, task5, task4);
+
+    List<Task> expected = Arrays.asList(task1, task3, task4, task5, task2);
+    List<Task> actual = new ArrayList<Task>();
+
+    TaskQueue taskQueue = new TaskQueue(taskList, HIGHEST_PRIORITY_FIRST);
+
+    for (int i = 0; i < expected.size(); i++) {
```
jamesnam-google: Same here

portfolio/src/test/java/com/google/sps/TaskQueueTest.java
```diff
@@ -73,4 +72,146 @@ public void checkCorrectSortingShortestTaskFirst() {
...
+            SchedulerTestUtil.PRIORITY_FIVE);
+
+    List<Task> taskList = Arrays.asList(task3, task1, task2, task5, task4, task6);
+
+    List<Task> expected = Arrays.asList(task6, task3, task5, task2, task4, task1);
+    List<Task> actual = new ArrayList<Task>();
+
+    TaskQueue taskQueue = new TaskQueue(taskList, HIGHEST_PRIORITY_FIRST);
+
+    for (int i = 0; i < expected.size(); i++) {
```
jamesnam-google: Same here

jamesnam-google (approved):

portfolio/src/test/java/com/google/sps/TaskQueueTest.java
raulcruise: Done.

portfolio/src/test/java/com/google/sps/TaskQueueTest.java
raulcruise: Done.

portfolio/src/test/java/com/google/sps/TaskQueueTest.java
raulcruise: Done.

portfolio/src/main/java/com/google/sps/data/HighestPriorityFirstScheduler.java
raulcruise: The reasoning is that I want to be able to index into currentAvailableTimes in the algorithm since I won't want to get the next availableTime from availableTimesGroup during each loop, so in some cases the index is increased and in some cases it remains the same. I had previously implemented the algorithm using the availableTimesGroup iterator but I had to use a bool to determine whether or not I would call next() on the iterator and it seemed a bit hacky to me.

portfolio/src/main/java/com/google/sps/data/HighestPriorityFirstScheduler.java
raulcruise: Done.

portfolio/src/main/java/com/google/sps/data/HighestPriorityFirstScheduler.java
raulcruise: I'll take any suggestions into using the iterator.

portfolio/src/main/java/com/google/sps/data/HighestPriorityFirstScheduler.java
raulcruise: Done.

portfolio/src/main/java/com/google/sps/data/HighestPriorityFirstScheduler.java
raulcruise: Done.

portfolio/src/main/java/com/google/sps/data/HighestPriorityFirstScheduler.java
raulcruise: Done.

portfolio/src/main/java/com/google/sps/data/HighestPriorityFirstScheduler.java
```diff
@@ -0,0 +1,139 @@
...
+
+      boolean doesTaskFit =
+          !currentScheduleTime
+              .plusSeconds(task.getDuration().getSeconds())
+              .isAfter(currentAvailableTimeRange.end());
+
+      if (doesTaskFit) {
+        ScheduledTask scheduledTask =
+            new ScheduledTask(
+                task, currentScheduleTime, /* isCompletelyScheduled = */ Optional.of(true));
```
hollyyuqizheng: After merging with the scheduling completeness enum PR, here I think we can change `Optional.of(true)` to `Optional.of(SchedulingCompleteness.COMPLETELY_SCHEDULED)`

portfolio/src/main/java/com/google/sps/data/HighestPriorityFirstScheduler.java
jamesnam-google: I think you don't have to use the iterator. But won't availableTimes be the same as currentAvailableTimes here? So maybe we can just assign currentAvailableTimes to availableTimes.

portfolio/src/main/java/com/google/sps/data/HighestPriorityFirstScheduler.java
jamesnam-google: Okay. Feel free to keep this logic.

portfolio/src/test/java/com/google/sps/TaskQueueTest.java
```diff
@@ -66,7 +65,153 @@ public void checkCorrectSortingShortestTaskFirst() {
...
+
+    List<Task> taskList = Arrays.asList(task3, task1, task2, task5, task4, task6);
+
+    List<Task> expected = Arrays.asList(task6, task3, task5, task2, task4, task1);
+    List<Task> actual = new ArrayList<Task>();
+
+    TaskQueue taskQueue = new TaskQueue(taskList, HIGHEST_PRIORITY_FIRST);
+
+    int i = 0;
+    while (i++ < expected.size()) {
```
jamesnam-google: We can introduce a method to taskQueue like isEmpty(), and then here do,

while (!taskQueue.isEmpty()) {

portfolio/src/test/java/com/google/sps/TaskQueueTest.java
```diff
@@ -66,7 +65,153 @@ public void checkCorrectSortingShortestTaskFirst() {
...
+
+    List<Task> taskList = Arrays.asList(task3, task1, task2, task5, task4);
+
+    List<Task> expected = Arrays.asList(task1, task3, task4, task5, task2);
+    List<Task> actual = new ArrayList<Task>();
+
+    TaskQueue taskQueue = new TaskQueue(taskList, HIGHEST_PRIORITY_FIRST);
+
+    int i = 0;
+    while (i++ < expected.size()) {
```
jamesnam-google: We can introduce a method to taskQueue like isEmpty(), and then here do,

while (!taskQueue.isEmpty()) {

portfolio/src/test/java/com/google/sps/TaskQueueTest.java
```diff
@@ -66,7 +65,153 @@ public void checkCorrectSortingShortestTaskFirst() {
...
+
+    List<Task> taskList = Arrays.asList(task3, task1, task2, task5, task4);
+
+    List<Task> expected = Arrays.asList(task2, task4, task5, task3, task1);
+    List<Task> actual = new ArrayList<Task>();
+
+    TaskQueue taskQueue = new TaskQueue(taskList, HIGHEST_PRIORITY_FIRST);
+
+    int i = 0;
+    while (i++ < expected.size()) {
```
jamesnam-google: We can introduce a method to taskQueue like isEmpty(), and then here do,

while (!taskQueue.isEmpty()) {

portfolio/src/test/java/com/google/sps/TaskQueueTest.java
```diff
@@ -66,7 +65,153 @@ public void checkCorrectSortingShortestTaskFirst() {
 
     TaskQueue taskQueue = new TaskQueue(taskList, SHORTEST_TASK_FIRST);
 
-    for (int i = 0; i < expected.size(); i++) {
+    int i = 0;
+    while (i++ < expected.size()) {
```
jamesnam-google: We can introduce a method to taskQueue like isEmpty(), and then here do,

while (!taskQueue.isEmpty()) {

jamesnam-google (approved):

portfolio/src/main/java/com/google/sps/data/HighestPriorityFirstScheduler.java
rgcase: I agree with James here, although I think it's fine to leave it as is now.

 The only thing is that with this logic I think you get results that are a little surprising. When I think "Highest Priority FIrst", I tend to think that it will try to schedule the highest priority tasks as soon as possible, but what's really happening (since you're also sorting the free time ranges by length) is that it's *choosing* the timeslot for the highest priority first, but not trying to put it into the first timeslot.

For example, if I have five 1-hour tasks, and I have a 5 hour block open at the beginning of the day and two 1-hour blocks later, I'll actually end up working on the 3rd, 4th, and 5th highest priority tasks first, then later work on the 1st and 2nd highest priority. Do I have that right? Is that what we're going for with this algorithm?

portfolio/src/main/java/com/google/sps/data/HighestPriorityFirstScheduler.java
raulcruise: Done.

portfolio/src/main/java/com/google/sps/data/HighestPriorityFirstScheduler.java
raulcruise: I am not sorting free time ranges by length. The only sorting that is done is sorting the Tasks by priority, then by duration and sorting the scheduledTasks by start time before returning them.

portfolio/src/test/java/com/google/sps/TaskQueueTest.java
raulcruise: Done.

portfolio/src/test/java/com/google/sps/TaskQueueTest.java
raulcruise: Done.

portfolio/src/test/java/com/google/sps/TaskQueueTest.java
raulcruise: Done.

portfolio/src/test/java/com/google/sps/TaskQueueTest.java
raulcruise: Done.

portfolio/src/main/java/com/google/sps/data/HighestPriorityFirstScheduler.java
rgcase: I see, sorry I misunderstood your comment above but that makes sense.

rgcase (approved):

PR #45, opened 2020-08-04
Use the Optional object's value instead of the object itself

hollyyuqizheng:
Because the Java object `isCompletelyScheduled` is of type Optional, the JavaScript cannot just use the ! check. Now the JavaScript line pulls out the value of the Optional object.

### DIFF
--- a/portfolio/src/main/webapp/request.js
+++ b/portfolio/src/main/webapp/request.js
@@ -32,6 +32,13 @@ const DESCRIPTION_INITIAL_TEXT = 'Description: ';
 const PRIORITY_INITIAL_TEXT = 'Priority: ';
 const INCOMPLETE_WARNING_TEXT = 'This task cannot be scheduled completely.'; 
 
+// Enums for scheduling completeness status
+const SCHEDULING_COMPLETENESS = {
+  NOT_SCHEDULED : 'NOT_SCHEDULED', 
+  PARTIALLY_SCHEDULED : 'PARTIALLY_SCHEDULED', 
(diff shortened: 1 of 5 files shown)


portfolio/src/main/webapp/request.js
```diff
@@ -188,7 +188,7 @@ function addScheduledTaskToDom(scheduledTask) {
   priorityText.setAttribute('data-task-priority', taskPriority);
   cardBody.appendChild(priorityText);
 
-  if (!isTaskCompletelyScheduled) {
+  if (!Object.values(isTaskCompletelyScheduled)[0]) {
```
jamesnam-google: This approach seems a bit hacky to me. One approach you can consider is to make the response contain two boolean values.

- Whether the task was scheduled at all (either or partial).
- Whether the task was completely scheduled.

Or, you can define some enum in both Java and Javascript that can represent these values.

ex:
public enum TaskSchedulingResult {
  NOT_SCHEDULED,
  PARTIALLY_SCHEDULED,
  COMPLETELY_SCHEDULED
}

Then you'll have to do some conversion from the optional to to this enum somewhere on the server side.

jamesnam-google (changes requested):

portfolio/src/main/webapp/request.js
hollyyuqizheng: I see. I added an enum in both Java and JavaScript for scheduling completeness status.

jamesnam-google (approved):

talvarez0507 (approved):

raulcruise (approved):

rgcase (approved):

PR #46, opened 2020-08-05
Add the Java Enum file for SchedulingCompleteness

hollyyuqizheng:
Using integer representation for the SchedulingCompleteness Enum.

### DIFF
--- a/portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java
+++ b/portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java
@@ -138,11 +138,11 @@ public class LongestTaskFirstScheduler implements TaskScheduler {
     // If iterating through all current time ranges finishes, and the task still isn't
     // completely scheduled, then this task is only partially scheduled.
     // Go through all the segments for this task, and set their completeness to partially scheduled.
-    Optional<Integer> schedulingCompletenessInt =
-        Optional.of(SchedulingCompleteness.PARTIALLY_SCHEDULED.getValue());
+    Optional<SchedulingCompleteness> schedulingCompleteness =
+        Optional.of(SchedulingCompleteness.PARTIALLY_SCHEDULED);
 
     for (ScheduledTask taskSegment : newScheduledTasks) {
-      taskSegment.setCompleteness(schedulingCompletenessInt);
+      taskSegment.setCompleteness(schedulingCompleteness);
     }
     return newScheduledTasks;
   }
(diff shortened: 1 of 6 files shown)


raulcruise (approved):

talvarez0507 (approved):

portfolio/src/main/java/com/google/sps/data/SchedulingCompleteness.java
```diff
@@ -0,0 +1,7 @@
+package com.google.sps.data;
+
+public enum SchedulingCompleteness {
```
jamesnam-google, line 7: You can consider assigning integer values, so that you don't have to perform string comparisons, and client can work with the int values.

ex:
public enum SchedulingCompleteness {
  NOT_SCHEDULED(0),
  ...(1),
  ...(2);

  private final int value;

  SchedulingCompleteness(int value) {
    this.value = value;
  }

  public int getValue() {
    return this.value;
  }
}

jamesnam-google (changes requested):

portfolio/src/main/java/com/google/sps/data/SchedulingCompleteness.java
hollyyuqizheng: I updated the enum accordingly. Right now, ScheduledTask takes in `Optional<Integer>` as the representation for scheduling completeness -- would this make sense? 
Another question I had was, based on the current implementation, I don't think the Optional object is necessary anymore? This completeness field will always have a value.

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java
```diff
@@ -138,11 +138,11 @@ public SchedulingAlgorithmType getSchedulingAlgorithmType() {
     // If iterating through all current time ranges finishes, and the task still isn't
     // completely scheduled, then this task is only partially scheduled.
     // Go through all the segments for this task, and set their completeness to partially scheduled.
-    Optional<SchedulingCompleteness> schedulingCompleteness =
-        Optional.of(SchedulingCompleteness.PARTIALLY_SCHEDULED);
+    Optional<Integer> schedulingCompletenessInt =
```
jamesnam-google: One thing you can consider doing is working with SchedulingCompleteness in most places (rather than Int) written in Java. For example, you can update ScheduledTask so that it takes in Optional<SchedulingCompleteness>, but when it does it sets another instance variable (schedulingCompletenessInt) of type Optional<Integer>. That way, you can work with SchedulingCompleteness which I think is more clear and less error-prone than Integer, but Javascript code can work with the integer value rather than string.

jamesnam-google (approved):

portfolio/src/main/java/com/google/sps/data/SchedulingCompleteness.java
jamesnam-google: I see. Yes I agree the optional is no longer necessary. Please see my comment below on using enum and integer.

portfolio/src/main/java/com/google/sps/data/LongestTaskFirstScheduler.java

portfolio/src/main/java/com/google/sps/data/ScheduledTask.java
```diff
@@ -11,6 +11,7 @@
   private final Task task;
   private final Instant startTime;
   private Optional<SchedulingCompleteness> schedulingCompleteness;
+  private Optional<Integer> schedulingCompletenessInt;
```
rgcase, line 14: Curious why we need this? Whenever you need to reference this int value can't you just call `schedulingCompleteness.get().getValue()`? Keeping them separate is more complicated and makes it easier to make a mistake where you get a mismatch between the `SchedulingCompleteness` and the `Integer`.

rgcase: Talked offline about this. I see now, it's because you want to serialize this object and send it to the client and have them work with the int value. Usually in this case you'd explicitly specify how you want the serialization happen so that the server only sees the enum and the client only sees the int, but maybe that's not straightforward with Gson. Okay that's fine you can leave it as is.

PR #47, opened 2020-08-06
Add Iterable interface to ArrayListTimeRangeGroup

hollyyuqizheng:
I make the TimeRangeGroup extend `Iterable<TimeRange>` so that forEach can be called directly on this class in the tests. I also migrated some test cases from ArrayListTimeRangeGroupTest to the main TimeRangeGroupTest, and I fixed a bug in the Linked List implementation after some new test cases failed.


raulcruise (approved):

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
```diff
@@ -5,7 +5,8 @@
 import java.util.List;
 
 /** Models an implementation of the TimeRangeGroup model using ArrayList. */
-public class ArrayListTimeRangeGroup extends AbstractListTimeRangeGroup implements TimeRangeGroup {
+public class ArrayListTimeRangeGroup extends AbstractListTimeRangeGroup
+    implements TimeRangeGroup, Iterable<TimeRange> {
```
jamesnam-google: I think TimeRangeGroup can extend Iterable<TimeRange>, as it exposes the iterator method.

portfolio/src/test/java/com/google/sps/ArrayListTimeRangeGroupTest.java
```diff
@@ -169,6 +169,34 @@ public void testAddOneOverlappingRange() {
     Assert.assertEquals(expected, actual);
   }
 
+  /** Testing for adding the exact same time range. */
+  @Test
+  public void testAddExactSameRange() {
```
jamesnam-google: I think we now have TimeRangeGroupTest [1] that tests both ArrayListTimeRangeGroup and LinkedListTimeRangeGroup. Can we add this test there, and maybe just remove ArrayListTimeRangeGroupTest.java entirely? But if you do so please make sure we port over any test that is in ArrayListTimeRangeGroupTest but not in TimeRangeGroupTest.

[1] https://github.com/googleinterns/step61-2020/blob/master/portfolio/src/test/java/com/google/sps/TimeRangeGroupTest.java

jamesnam-google (changes requested):

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
hollyyuqizheng: I updated the abstract class to implement Iterable<TimeRange>.

portfolio/src/test/java/com/google/sps/ArrayListTimeRangeGroupTest.java
hollyyuqizheng: Updated.

portfolio/src/main/java/com/google/sps/data/ArrayListTimeRangeGroup.java
jamesnam-google: I would suggest that you make the interface extend iterable, because often you are working with interface and not the impl class. Here are some examples where the interface extends iterable.

https://docs.oracle.com/javase/8/docs/api/java/util/List.html
https://docs.oracle.com/javase/7/docs/api/java/util/Set.html

portfolio/src/main/java/com/google/sps/data/LinkedListTimeRangeGroup.java
```diff
@@ -83,8 +83,17 @@ public void addTimeRange(TimeRange timeRange) {
       // the same element then there is no need to add it. However, if
       // lastExaminedTimeRange does not match, this means lastExaminedTimeRange
       // actually belongs at the end of the new list.
+      //
```
jamesnam-google: Unnecessary line

portfolio/src/main/java/com/google/sps/data/LinkedListTimeRangeGroup.java
```diff
@@ -83,8 +83,17 @@ public void addTimeRange(TimeRange timeRange) {
...
+      // Note: we do not use TimeRange's equals() method here but compare the objects directly.
+      // If currentRange and lastExainedRange both point to the same time range object,
```
jamesnam-google: lastExainedRange => lastExaminedRange

jamesnam-google (approved):

talvarez0507 (approved):

rgcase (approved):

